PR #10, opened 2018-08-30
Add user management functions for super users

Fingel:


tom_common/views.py
```diff
@@ -1,3 +1,53 @@
...
+class UserDeleteView(DeleteView):
+    success_url = reverse_lazy('user-list')
+    model = User
+
+    @method_decorator(user_passes_test(lambda u: u.is_superuser))
+    def dispatch(self, *args, **kwargs):
+        return super().dispatch(*args, **kwargs)
+
+
+class TOMPasswordChangeView(FormView):
```
dmcollom: Not sure why the naming convention is so radically different here?

Fingel: You're right, silly. I'll change it.

PR #14, opened 2018-09-24
Feature/observing module

Fingel:


tom_observations/facilities/lco.py
```diff
@@ -0,0 +1,294 @@
...
+                        }
+                    ],
+                    "windows": [
+                        {
+                            "start": self.cleaned_data['start'],
+                            "end": self.cleaned_data['end']
+                        }
+                    ],
+                    "location": {
+                        "telescope_class": "1m0"
```
dmcollom: Do we want to open Pandora's box and add telescope support, or are we leaving this as-is?

PR #16, opened 2018-09-28
Feature/install script

Fingel:


tom_setup/management/commands/tom_setup.py
```diff
@@ -0,0 +1,117 @@
...
+        if major == 3 and minor == 6:
+            try:
+                import dataclasses  # noqa
+            except ImportError:
+                self.exit('Could not load dataclasses. Please use Python >= 3.7 or 3.6 with dataclasses installed')
+        elif major < 3 or minor < 7:
+                self.exit('Incompatible Python version found. Please install Python >= 3.7')
+        self.ok()
+
+    def create_project_dirs(self):
```
dmcollom: Do we want to add a custom exception for any other cases? Lacking permissions or other? Or just let Python exceptions suffice?

tom_setup/management/commands/tom_setup.py
```diff
@@ -0,0 +1,117 @@
...
+    def ok(self):
+        self.stdout.write(self.style.SUCCESS('OK'))
+
+    def status(self, msg):
+        self.stdout.write(msg, ending='')
+        sys.stdout.flush()
+
+    def welcome_banner(self):
+        welcome_text = (
+            'Welcome to the tom_setup helper script. This will help you get started with a new tom.\n'
```
dmcollom: Super, incredibly nitpicky but let's capitalize TOM since it's an acronym. Branding and all that.

tom_setup/management/commands/tom_setup.py
```diff
@@ -0,0 +1,117 @@
...
+        except FileExistsError:
+            pass
+        self.ok()
+
+    def run_migrations(self):
+        self.status('Running initial migrations... ')
+        call_command('migrate', verbosity=0, interactive=False)
+        self.ok()
+
+    def get_target_type(self):
```
dmcollom: Is there a corresponding change somewhere else to only display forms and fields corresponding with the selected target type, or are we not considering that part of this story?

tom_setup/management/commands/tom_setup.py
Fingel: I'm not sure what other exceptions could happen at this point, if they've already installed django and are running this script then that already eliminates a lot of error cases. If anything pops up we can add more handling.

tom_setup/management/commands/tom_setup.py
Fingel: I disagree. `tom_setup` is the name of the command. All management commands are lower case, as they map to python files, which are also lower case (tom_setup.py)

tom_setup/management/commands/tom_setup.py
Fingel: Nope, not yet. But that's the plan.

tom_setup/management/commands/tom_setup.py
dmcollom: I meant the "get started with a new tom" portion.

Fingel: Ah, you're right. I'll fix that.

Issue #46, opened 2018-10-25
Implement code hooks


Fingel:
There are certain actions that happen within the TOM toolkit which should trigger hooks: other code that gets run. This is important for specific science use cases that are to be implemented by the TOM author. An example would be: submitting the results of an observation to a 3rd party server when an observation becomes complete.

Some places where code hooks should reside:

* Observation status changes
* Adding a target
* Submitting an observation

PR #47, opened 2018-10-26
Add support for external code hooks

Fingel:
Adds support for providing arbitrary code hooks to certain actions
within the TOM toolkit. The goal of this is to make it easy for TOM
authors to run custom scripts without needing to change the TOM
internals.

Documentation: https://tomtoolkit.github.io/docs/custom_code

Fixes https://github.com/TOMToolkit/tom_base/issues/46

### DIFF
--- a/tom_base/settings.py
+++ b/tom_base/settings.py
@@ -167,6 +167,11 @@ FACILITIES = {
     }
 }
 
+HOOKS = {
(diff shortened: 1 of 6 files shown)


dmcollom:
Looks like tests need to be updated but otherwise looks good.

PR #50, opened 2018-11-05
Feature/image display

dmcollom:
Adds support for light curves on target detail page
Displaying fits files as PNGs on target detail, observation record, and data product list
Adds support for selecting a featured fits file and light curve
Restructures target detail page


tom_observations/models.py
```diff
@@ -70,10 +94,58 @@ class DataProduct(models.Model):
...
+    featured = models.BooleanField(default=False)
 
     class Meta:
         ordering = ('-created',)
+        get_latest_by = ('modified',)
 
     def __str__(self):
         return self.data.name
+
+    def get_file_name(self):
```
Fingel: Consider using https://docs.python.org/3/library/os.path.html#os.path.basename
It doesn't rely on the OS using '/' as a path separator.

tom_observations/models.py
```diff
@@ -70,10 +94,58 @@ class DataProduct(models.Model):
...
+        return self.data.name.split('/')[-1]
+
+    def get_file_extension(self):
```
Fingel: https://docs.python.org/3/library/os.path.html#os.path.splitext

tom_observations/models.py
```diff
@@ -70,10 +94,58 @@ class DataProduct(models.Model):
...
+    def get_png_data(self, min_scale=40, max_scale=99):
+        path = settings.MEDIA_ROOT + '/' + str(self.data)
+        image_data = fits.getdata(path, 0)
+        fig = plt.figure()
+        vmin = 0
+        vmax = 0
+        if image_data.size > 0:
+            vmin = np.percentile(image_data, min_scale)
+            vmax = np.percentile(image_data, max_scale)
+        plt.imshow(image_data, vmin=vmin, vmax=vmax)
```
Fingel: We probably don't want to be trying to show the image from the server side.

tom_observations/models.py
```diff
@@ -70,10 +94,58 @@ class DataProduct(models.Model):
...
+
+    def get_file_name(self):
+        return self.data.name.split('/')[-1]
+
+    def get_file_extension(self):
+        name = self.data.name.partition('.')
+        extension = name[2] if len(name)==3 else ''
+        return extension
+
+    def get_light_curve(self, error_limit=None):
```
Fingel: I'm not going to pretend like I fully understand what's going on here. But a few questions:
1. What happens if you run this on a non-fits file?
2. Might this be better placed in it's own module and then referenced here?
3. Config? I'm guessing sci people might want to tweak scaling and such to math their data.

tom_observations/models.py
dmcollom: Good call--done.

tom_observations/models.py
dmcollom: Also done.

tom_observations/models.py
dmcollom: 1. So this actually runs on any simple four-column delimited file (not fits files), but there IS an expected format. As the scientists have mentioned, the expectation, I believe, is that the users write code to transform it properly. We could either document the format thoroughly or it could require column headers that would need to be processed. Running it on a different file type would, presumably, crash, but why would you call get_light_curve on a file that doesn't contain light curve data?
2. Yes, probably belongs elsewhere. I'll move it.
3. My understanding was that Rachel wanted a way to display light curves at a minimum, and while we can add configuration options, I'm not sure what those would be, nor what would be in-scope or out-of-scope for this.

tom_observations/models.py
dmcollom: Good point--I'll do my best to get this working without that, but I'm not sure how.

dmcollom:
Re: data product tagging, there isn't a way at present, other than at upload. We could either (or both) have a default tag for the various archive files, and/or add a manual tagging option, though where that belongs in the UI I don't know.

I can work on caching.

tom_observations/models.py
dmcollom: Actually very easy, this is updated.

dmcollom:
Caching has been added, light curve has been moved, and tags can be updated

tom_base/settings.py
```diff
@@ -137,6 +137,17 @@
 DATE_FORMAT = 'Y-m-d'
 
 
+# Caching
+# https://docs.djangoproject.com/en/dev/topics/cache/#filesystem-caching
+
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
+        'LOCATION': os.path.join(BASE_DIR, 'tmp')
```
Fingel: I meant use the OS temp file directory. Once again, python has an answer:
https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

Issue #40, opened 2018-10-23
Add MPC catalog harvester


Fingel:
Users should be able to ingest targets from the MPC via a TOM catalog harvester. We can achieve this thanks to the good work of @dmcollom with astroquery.

PR #52, opened 2018-11-07
App MPC harvester class

Fingel:
Adds MPC as an available source for querying targets

Fixes https://github.com/TOMToolkit/tom_base/issues/40

### DIFF
--- a/tom_catalogs/harvesters/mpc.py
+++ b/tom_catalogs/harvesters/mpc.py
@@ -1,0 +1,26 @@
+from tom_catalogs.harvester import AbstractHarvester
+
+from astroquery.mpc import MPC
+
+
+class MPCHarvester(AbstractHarvester):
+    name = 'MPC'
+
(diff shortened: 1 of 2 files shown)


tom_catalogs/harvesters/mpc.py
```diff
@@ -0,0 +1,26 @@
+from tom_catalogs.harvester import AbstractHarvester
+
+from astroquery.mpc import MPC
+
+
+class MPCHarvester(AbstractHarvester):
+    name = 'MPC'
+
+    def query(self, term):
+        self.catalog_data = MPC.query_object('asteroid', name=term)
```
dmcollom: Do we want to add support for querying comets as well?

Fingel: I'm not sure how we would, because the search form only allows specifying the search term. So punting on this for now.

PR #62, opened 2019-01-02
Feature/gemini url

bryanmiller:
The initial Gemini facility module to use the legacy Gemini URL triggering.


bryanmiller:
Note, this requires the installation of https://github.com/bryanmiller/gsselect, which requires least numpy 1.15.4.

bryanmiller:
Gsselect was my first attempt at making a pip package so I'm still learning. I updated gsselect and I think that it will work correctly now. This meant updating gemini.py so there is a new pull request.

PR #64, opened 2019-01-04
Feature/gemini url

bryanmiller:
Here is a new pull request for the Gemini facility integration using the latest gsselect package reorganization.


Fingel:
Hi Bryan! Looks and works great. I've made some changes I'll put into a different PR so David can take a look.

PR #68, opened 2019-01-17
Add Scout neo module

Fingel:


tom_alerts/brokers/scout.py
```diff
@@ -0,0 +1,79 @@
...
+from tom_targets.models import Target, TargetExtra
+
+SCOUT_URL = 'https://ssd-api.jpl.nasa.gov/scout.api'
+
+
+class ScoutQueryForm(GenericQueryForm):
+    pass
+
+
+def hours_min_to_decimal(val):
```
dmcollom: We should probably just use Astropy's conversion tools instead, which are already being leveraged in the visibility calculations.

Fingel: Good idea

PR #73, opened 2019-01-24
Inclusion tag for target distribution plot

Fingel:
Adds an inclustion tag that plots supplied targets by ra/dec using
plotly. Where this plot should be places can be discussed but for now it
is above the target list table and plots the targets that match the
supplied filters.


Fingel:
@jfrostburke @cmccully Should the center of the RA axis be 0 or 180?

Fingel:
Cool, this should be good then.

Fingel:
We'll see how this plays out: https://community.plot.ly/t/tick-labels-for-geo-plots/18771

PR #75, opened 2019-01-29
Feature/time series support

dmcollom:


tom_reduced_data/models.py
```diff
@@ -0,0 +1,33 @@
...
+        verbose_name='Datum Source Location',
+        help_text='URL or path to original target source reference'
+    )
+
+
+class ReducedDatum(models.Model):
+    source = models.ForeignKey(ReducedDatumSource, null=False, on_delete=models.CASCADE)
+    target = models.ForeignKey(Target, null=False, on_delete=models.CASCADE)
+    data_product = models.ForeignKey(DataProduct, null=True, on_delete=models.CASCADE)
+    data_type = models.TextField(
```
Fingel: TextFields should be avoided as much as possible: they are more expensive than CharFields and are usually used for long form text where you don't know ahead of time how big they will be.

tom_reduced_data/models.py
```diff
@@ -0,0 +1,33 @@
...
+        max_length=100,
+        choices=getattr(settings, 'DATA_TYPES', ('', ''))
+    )
+    timestamp = models.TextField(null=False, blank=False, db_index=True)
```
Fingel: Timestamp is a TextField? If there is a good reason it's definitely worthy of a comment.

tom_reduced_data/models.py
```diff
@@ -0,0 +1,33 @@
...
+    value = models.FloatField(null=False, blank=False)
+    label = models.TextField(null=True)
```
Fingel: Don't use Null for text/charfields, default to the empty string so we only have 1 way to represent the empty value.

tom_reduced_data/models.py
```diff
@@ -0,0 +1,33 @@
+from django.db import models
```
Fingel: Curious to know why you chose this to be a new app instead of using tom_dataproducts, for example.

tom_alerts/brokers/mars.py
```diff
@@ -187,22 +188,53 @@ def fetch_alerts(clazz, parameters):
...
+            alerts = [alert]
+        else:
+            try:
+                alert_sources = ReducedDatumSource.objects.filter(
+                    id__in=ReducedDatum.objects.filter(target=target).values_list('source').distinct(),
+                    name=self.name
+                )
+                for alert_source in alert_sources:
+                    alerts.append(self.fetch_alert(alert_source.location))
+            except ObjectDoesNotExist:
```
Fingel: If this is mean to catch exceptions from the filter at line 207, I don't think it will work. .get() would raise ReducedDatumSource.DoesNotExist, but using .filter() will just return an empty list.

tom_alerts/alerts.py
```diff
@@ -88,3 +89,46 @@ def save(self, query_id=None):
         query.parameters = self.serialize_parameters()
         query.save()
         return query
+
+
+class GenericBroker(ABC):
+
+    @abstractmethod
+    def fetch_alerts(parameters):
```
Fingel: Are these not instance methods? Prob need the self parameter.

tom_reduced_data/models.py
dmcollom: I think it was hedging that the complexity wouldn't drive it out of the scope of tom_dataproducts, though it seems that concern wasn't warranted. I can move it back with likely very little effort. Probably makes more sense there, would you agree?

tom_alerts/alerts.py
dmcollom: This has been corrected in the most recent push.

tom_alerts/brokers/mars.py
dmcollom: This has been corrected in the most recent push.

tom_reduced_data/models.py
dmcollom: This has been corrected in the most recent push.

tom_reduced_data/models.py
dmcollom: This has been corrected in the most recent push.

dmcollom:
Just let me know on rolling it into the tom_dataproducts app and I'll make the required refactor. Otherwise all the other concerns have been addressed.

tom_reduced_data/models.py
Fingel: I'd still argue that using a real timestamp is better. You can always translate from one time format to a timezone. Using a float leaves it ambiguous, what does it mean? Seconds since unix epoch? Milliseconds? Hours since the big bang? Days since I've properly bathed?

tom_reduced_data/models.py
Fingel: I think it does. Not a big deal either way.

tom_reduced_data/models.py
dmcollom: Not sure float values go that high.

Your point is taken, though. Genuinely not sure why I blanked on the existence of a DateTimeField. It's now correct.

tom_reduced_data/models.py
dmcollom: It's all been moved.

PR #78, opened 2019-01-31
Fixed crash resulting from non-detections from MARS not having magpsf

dmcollom:
Modified JD conversion to be timezone-aware to stop Django from complaining

### DIFF
--- a/tom_alerts/brokers/mars.py
+++ b/tom_alerts/brokers/mars.py
@@ -4,7 +4,7 @@ from urllib.parse import urlencode
 from dateutil.parser import parse
 from django import forms
 from crispy_forms.layout import Layout, Div, Fieldset, HTML
-from astropy.time import Time
+from astropy.time import Time, TimezoneInfo
 
 from tom_alerts.alerts import GenericQueryForm, GenericAlert
 from tom_targets.models import Target, TargetExtra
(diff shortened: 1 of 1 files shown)


tom_alerts/brokers/mars.py
```diff
@@ -217,15 +217,17 @@ def process_reduced_data(self, target, alert=None):
...
-                jd = Time(prv_candidate['candidate']['jd'], format='jd')
-                magnitude = prv_candidate['candidate']['magpsf']
-                rd, created = ReducedDatum.objects.get_or_create(
-                    timestamp=jd.datetime,
-                    value=magnitude,
-                    source=reduced_datum_source,
-                    data_type='PHOTOMETRY',
-                    target=target)
-                rd.save()
+                if prv_candidate['candidate'].get('jd', False) and prv_candidate['candidate'].get('magpsf', False):
```
Fingel: No need to use .get(x, False) .get() returns None if the key is not found so falsey anyway.

Fingel: Also you could just use `if all([key in prv_candidate['candidate'] for key in ['jd', 'magpsf']]):`

dmcollom: Good call. Updated.

PR #79, opened 2019-02-01
Closes the loop on issue #22

### LINKED ISSUE #22: Broker alerts: bulk target creation.
The result list for broker queries should allow for bulk creation of targets. The user should be able to select multiple (or all) returned alerts and create targets from them at once.

dmcollom:
Fixes crash when submitting multiselect forms with no selections in dataproducts and alerts
Adds saner redirection when creating a target from a single alert

Fixes #22


Fingel:
Test?

PR #81, opened 2019-02-12
Feature/js9

Fingel:


jfrostburke:
Tested on a local copy and looks good. The "light window" thing is kind of weird but astronomers' familiarity with ds9 makes it work for me.

PR #82, opened 2019-02-13
Feature/easy extra fields

Fingel:
This allows a declarative way of adding custom fields to targets. It still uses the underlying key/value model but the defined fields should look and act the same as the normal model fields.

This should work for the "adding redshift" use case from Snex 2.0 without the need of copying entire models.


Fingel:
This is crap

PR #83, opened 2019-02-13
Feature/spectra plotting

dmcollom:


Fingel:
Why can't we change tags anymore?

dmcollom:
Changing tags just doesn't seem to have a use case at the moment--for spectra and light curves, we're ingesting the data on upload and displaying it from the database rather than the file, and so changing the tag would be purely cosmetic. For fits files, they'll likely mostly come straight from archives, and the only reason you'd change it is if it were mislabeled in the first place, so it's no less cumbersome to just delete and re-upload. Happy to change my mind on that though.

Fingel:
Deleting and reuploading could be pretty cumbersome if it's a large file, especially if we have ambiguous  tags like image vs fits file. But that's fine, we can leave it out for now.

PR #90, opened 2019-03-07
Feature/scheduled times

Fingel:
Changes the status method of facilities to a dictionary that also returns the scheduled window for an observation (if any). Displays these days in various places: observation list, observation list for target, and adds a notification to the target detail page.


tom_targets/templates/tom_targets/target_detail.html
```diff
@@ -9,6 +9,9 @@
   <div class="col-md-4">
     <div id="target-info">
       {% target_feature object %}
+      {% if object.future_observations %}
+      <div class="alert alert-success">{{ object.future_observations|length }} upcoming observations</div>
```
dmcollom: I'd use the django pluralize tag here, just to avoid that minor pitfall.

Fingel: good call

PR #93, opened 2019-04-04
Added support for email notifications

rachel3834:
Obviously this is my first contribution here, so I have attempted to integrate this function in a logical way but I'm open to suggestions if you think there is a better one.


Fingel:
Hi Rachel,
I think the code in the pull request is good. I think it should be accompanied by some documentation about how to actually configure the email settings afterwards. I'll write up some documentation for it and then we can merge this. Thanks!

PR #94, opened 2019-04-09
Adding JPEGs

zemogle:
I've added JPEGs as a new `DataProduct` type. I've written methods to produce JPEGs in the base code and added a setting `AUTO_THUMBNAILS` (defaults to `False`) which will enable the creation of thumbnails when a FITS type data product is saved.

I've made the default image size a setting `THUMBNAIL_MAX_SIZE`. Its currently a blank tuple which will mean the code will give the maximum image resolution.

This code does require the `astropy` branch of `fits_to_image`. If you're happy for me to add this to pypi we can make it a requirement.


Fingel:
Thanks for the PR!
I'm still a bit confused about this. @dmcollom already added code to generate thumbnails from uploaded fits files so it's not clear if the intent is to:
1) replace that functionality (which this PR doesn't seem to do)
2) augment it (in this case why the separate image type?)
I think you and David should probably have a conversation about how to proceed.

zemogle:
1. Auto-generated thumbnails are fine for quick inspection but it is really useful to have actual files for the education TOM (e.g. 1 of the features I'll be adding is timelapse creation). Many applications of the TOM Toolkit I've talked to partners about, want a page with thumbnails of all their data. This would be tedious (and slow) to autogenerate each time the page is loaded.

1. If we have actual files, they should be dataproducts.

1. The current code doesn't work for me (and is totally undocumented). When I eventually found it I presumed this was either a placeholder or something which had been ditched. There is code in place for something to happen on e.g. the dataproduct list page https://github.com/TOMToolkit/tom_base/blob/c22e30152d810e3891df86c0cc80a8f3b6c0ebee/tom_dataproducts/templates/tom_dataproducts/dataproduct_list.html#L50,  the observation record detail view  https://github.com/TOMToolkit/tom_base/blob/c22e30152d810e3891df86c0cc80a8f3b6c0ebee/tom_observations/views.py#L161
Neither of those examples produce anything on my sandbox until you change the `tag` to `Fits File` which can only be done from the Django admin. Then I can get a thumbnail by featuring the file but it still doesn't appear on  `dataproduct_list.html`

1. I really don't understand why you'd use MatPlotlib to make an image instead of PIL/pillow. Particularly as `fits2image` is a really well tested, flexible, light-weight and mature package.

1. Rainbow colour scale is totally unphysical, greyscale is much more useful.

The thing that worries me most is why, if you both knew this code existed didn't you point me to that instead of letting me write some new code?  @dmcollom has been totally silent about adding this feature on Slack and has expressed almost no interest in anything else I've put on Slack.

Fingel:
Hi @zemogle,

Sorry for the misunderstanding. This is good code, and I agree that using fits2image is better than using matplotlib. My concern is that I don't think we need to have both methods in place, it's confusing. I'd like @dmcollom to work with you to incorporate what you've done in a more complete manner.

dmcollom:
Hi Edward,

First of all, my apologies if you feel I haven't been attentive enough--I'll do my best to be more communicative in the future. I simply entirely blanked on this--we had originally supported PNGs and [switched to JPEGs for thumbnails later](https://github.com/TOMToolkit/tom_base/commits/3e546ab7ee4abcb7b533c0804bd9a935429fa46f/tom_observations/models.py), so from my perspective, once we said to update the definition, it was a fairly simple change, but we should probably do better about giving starting points to look at and documenting better. I _do_ think it's a bit disingenuous and rather unhelpful to say that I've expressed no interest in anything you've put on Slack; it takes a very short scroll up the channel to see that we've engaged on a number of things in the past, I recently corrected the non-sidereal issues you brought up, and in fact, I was the one that added you to the channel in response to your initial set of questions.

Using matplotlib was a product of me asking another team member what the best way to display fits images was, and not being aware of the existence of fits2image as a newer team member.

You are correct that that was part of functionality that has evolved/been partially removed and was poorly handled since then. I think it's important to remember that, while a situation like that should have been handled better, we're still trying to figure out what's going to provide value and we really aren't 100% sure how certain features are going to be used and how their usefulness can be maximized. It's why we appreciate having alpha users so much--we weren't aware of greyscale being the more useful colorscale. However, to your point about performance, we cache thumbnails--that said, pre-generated thumbnails are a more efficient solution. Ours was put in place to support both fits and image files, and I chose to do that without creating a second file for a thumbnail.

Overall, if we aren't concerned about the additional storage (and I'm not), I think the generated thumbnails is superior and probably should replace the existing functionality. I'll pull this down and test/review it, and once you merge it there's a good amount of cleanup that we should do.

tom_dataproducts/views.py
```diff
@@ -23,6 +23,10 @@
 from tom_observations.facility import get_service_class
 
 
+class DataProductJpegCreate(LoginRequiredMixin, View):
```
dmcollom: It doesn't look like this is ever used, and isn't referred to in urls.py--can it be removed?

tom_dataproducts/utils.py
```diff
@@ -1,11 +1,17 @@
...
+import shutil
+import tempfile
 
+from astropy.io import fits
 from astropy.time import Time
 from datetime import datetime
-from .models import ReducedDatum
 from django.conf import settings
+from django.core.files import File
+from fits2image.conversions import fits_to_jpg
```
dmcollom: It looks like the fits2image dependency wasn't added to setup.py -- @Fingel uploaded it to PyPi, so just make sure you add it.

tom_dataproducts/views.py
zemogle: Good call.

tom_dataproducts/utils.py
zemogle: Done

zemogle:
Thanks for your reviews @dmcollom. It would be great if you did want switch over to this code. Using `fits2image` will be really useful for making colour images too.

I've fixed up a couple of the points but as you've suggested there is some additional tidying needed in my code. A couple of things strike me as being useful:

- How do we want to replace the current image handling in templates? I'm happy to think about this more.
- If we always make thumbnails, then these should be the only type of file which are "Featured"
- I've allowed for specifying a default size for the thumbnails, where the default is max size. We might want to create small thumbnails as a default and have a different way of letting the user enable full size thumbnails.
- `fits2image` has lots of options. The median and percentile settings will be the most useful to astronomers, we might want to allow those to be globally set in `settings.py`

dmcollom:
- I'm fine with leveraging this code and using the most recent thumbnail by default for featured images--however, if that's the case, `AUTO_THUMBNAIL` should default to `True` and it should be added as a configuration option in the setup script.
- I'd be curious if you think there's any use case to even supporting JPEG or PNG upload at this point--which would mean excluding `IMAGE_FILE` from the user-facing data product type selection if not. Image files were very much an early effort to have something non-text-based, and may not be a useful feature anymore.
- If thumbnails will be the featured image by default, I would think the default size should be reasonably sized for the current target detail page, not smaller. Perhaps that should be included in the setup script as well.
- I will defer to you on `fits2image` options, but I certainly would endorse giving more (documented) settings in `settings.py`.

zemogle:
* I'd be happy with `AUTO_THUMBNAIL` defaulting to `True` but not sure how this could be made a configuration option in the setup script. Is that something you could work on @dmcollom ?

* I can see that people might still want to upload a JPEG/PNG, for an annotated image. This is sometimes done for finder charts, particularly if you don't want your target in the middle of the frame. Should this be a different sort of `IMAGE_FILE`?

* Good point. I was thinking a full resolution image (4k x 4k for 1m's) might be overkill. Maybe leaving the defaults as (1k x 1k)?

* I'll add a way to make the thumbnail settings configurable with sensible defaults and update the PR

dmcollom:
As much as I don't want to add another data product type, if we're going to distinguish thumbnails, perhaps we should have a `THUMBNAIL` data product type in addition to the `IMAGE_FILE`, with the distinction that a `THUMBNAIL` is generated.

That seems like a reasonable default size.

As far as the setup script goes, that was mostly under the purview of @Fingel but based on this snippet it should be pretty cut and dried:

`def get_target_type(self):
        prompt = 'Which target type will your project use? {}'.format(self.style.WARNING('[SIDEREAL/NON_SIDEREAL] '))
        target_type = input(prompt).upper()
        if target_type not in ['SIDEREAL', 'NON_SIDEREAL']:
            self.stdout.write('Error: invalid type {} valid types are SIDEREAL, NON_SIDEREAL'.format(target_type))
            self.get_target_type()
        self.context['TARGET_TYPE'] = target_type`

I can do it after you merge your PR if you'd rather I take care of it.

Fingel:
Hi,

How about this approach:

Let `IMAGE_FILE` be it's own data product type.

Implement a `get_preview(self, size=None)` function for dataproducts. For both fits files and images it can return a thumbnail sized image of the dataproduct (perhaps by default). It also takes an argument for size, which determines the size of the returned image. 

For non image data products, we can decide what to return for `get_preview()`. It could even just be a static image of a filetype icon or something for PDFs or whatever.

We'd probably want to run the `get_preview` function for thumbnail sizes as soon as possible to make sure it's cached.

Does that sound reasonable?

dmcollom:
I'm very much on board with that--I think it's a great idea. I would say call `get_preview` when the DataProduct` is created--there's already some (possibly obsolete) caching code in `DataProductFeatureView` for a quick example.

zemogle:
Thanks for both of your comments (been out for a few days for Easter). I really like the direction this is going. We can then use these thumbnails on the data and observation record listing pages.

I'm happy to work on that and add it to this PR.

@dmcollom if you could take care of updating the setup script that would be a big help. 

Thanks!

Fingel:
Has there been any progress on this? It'd be great to get this merged.

zemogle:
Yes. I've made good progress on this. I added a `get_preview` def on the `DataProduct` model. It is intended to either serve a thumbnail url or generate then serve a thumbnail url.

Something that I would like your advice on is how we tell an image dataproduct type from a doc dataproduct type.

tom_dataproducts/models.py
```diff
@@ -75,22 +79,24 @@ def get_file_name(self):
...
-            plt.axis('off')
-            ax = plt.gca()
-            ax.xaxis.set_major_locator(matplotlib.ticker.NullLocator())
-            ax.yaxis.set_major_locator(matplotlib.ticker.NullLocator())
-            plt.imsave(buffer, image_data, format='jpeg')
-            buffer.seek(0)
-            plt.close(fig)
-        return b64encode(buffer.read()).decode('utf-8')
+    def get_preview(self, size=settings.THUMBNAIL_DEFAULT_SIZE, redraw=False):
+        from .utils import create_jpeg
```
dmcollom: I know that this is imported here to avoid a circular import, but should create_jpeg() simply live here instead? It takes the data product as the first parameter anyway, perhaps it should just be a method on the model.

tom_dataproducts/utils.py
```diff
@@ -45,3 +51,48 @@ def process_data_product(data_product, target):
             timestamp=datetime.now(),
             value=json.dumps(spectrum)
         )
+
+def create_image_dataproduct(data_product):
```
dmcollom: Can we maybe just call this create_thumbnail? Seems to be a clearer designation.

zemogle: This function is creating a `DataProduct` whereas `create_jpeg` is doing the thumbnail creation. I think the current def name is probably clearer.

tom_dataproducts/models.py
zemogle: Good idea. It actually makes everything much simpler if that's part of the model class.

Fingel:
Can we switch the base of this PR to `development`?

zemogle:
This should fix the outstanding issues on this PR, thanks to @joesingo !

dmcollom:
Looks good! I fixed merge conflicts resulting from Tuesday's changes and fixed a really small issue in the call to create_thumbnail, but otherwise it should be fine to merge.

PR #102, opened 2019-06-05
Add a kwarg to target save for saving extras

Fingel:
Adds a programmatic method for saving target extras. The `extras` kwarg
can be passsed to a Target's `save()` method, allowing for the user to
pass in a dictionary of key/value pairs that will be saved as
TargetExtras for that target.


Fingel:
I think we should merge development and make a release once this is merged.

PR #110, opened 2019-06-28
Feature/observation plot

Ara-yjx:
Add geo plot to observation page.
Color is based on observation status:
- Green if it is a terminal status provided by the facility
- Red if not
- Gray if status is an empty string

Implement get_wavelength_units and get_flux_constant methods in GEMFacility


tom_observations/templatetags/observation_extras.py
```diff
@@ -19,3 +23,89 @@ def observation_list(target=None):
     else:
         observations = ObservationRecord.objects.all().order_by('-created')
     return {'observations': observations}
+
+@register.inclusion_tag('tom_observations/partials/observation_distribution.html')
+def observation_distribution(observations):
+
+    # this does not work
```
dmcollom: Can we either remove or simplify this comment block? I'd rather not clutter the code with what doesn't work.

Fingel:
Merge?

tom_observations/facilities/gemini.py
```diff
@@ -500,6 +500,14 @@ def get_observation_url(clz, observation_id):
         # return PORTAL_URL + '/requests/' + observation_id
         return ''
 
+    @classmethod
```
Fingel: This isn't necessary anymore. You can remove these methods unless Gemini actually has those units.

PR #113, opened 2019-07-02
Feature/export target data

Ara-yjx:
Add "export all targets" and "export filtered targets" button on main page;
Enhance "import targets" function so that it deals with blank data stably;


Ara-yjx:
Hi Austin! Iris and my initial purpose for this approach is to archive the file and meanwhile make downloading simpler. But we agree with you that we should do it just in memory. We will fix this immediately. Thanks!

> Hi Ara! Thanks for the help on this, we're super happy to have you onboard.
> I was just curious the reasoning for saving the files locally, instead of just returning them to the browser. Did you specifically want to save the files locally or was that just the easiest way to make the download work? You can create files entirely in memory without writing them to disk, David or I can help with that if you'd like.

dmcollom:
@Ara-yjx thanks for the changes, just the last thing before I merge this in is a couple of rogue print statements.

PR #116, opened 2019-07-09
Added TargetList (aka target grouping) functionality

xuiris:
Users can now group targets (linked with model TargetList). Makes management of targets easier for users.
- New dropdown under "Target" to view groups called "Target Grouping"
  - Groupings listed/created/deleted here 
- Target listing page updated:
  - allows for multi adding/removal of targets to existing groups
  - ability to select targets on page OR select all listed targets
  - allows filtering by group
  - all filters persist after adding/removing
- Target details page has new "Manage Groups" tab for viewing groups the target is part of, as well as the ability to remove it from its groups.

Unit tests were added.

Collab with @Ara-yjx


dmcollom:
Really like how this was done, and I think it's pretty easy to navigate for the user. A few things to address though:

- It's small, but removing a target from a list it doesn't belong to just says "success". It may not really matter, but I think I'd prefer using messages.warning with something along the lines of "target {name} doesn't belong to that list".
- I think failed targets should appear in a messages.error rather than warning. Also, you can add the failed ones to a messages.error and the successful ones to a messages.success--it'll simply produce two banners.
- Looks like removing or adding a target to a list resets the filters--those should probably persist.
- Unit tests!! Never a wrong time to write them.

Fingel:
I agree with David, nice work!

Another useful addition might be a "select all" checkbox at the top of the target table, that way the user has the option to add bulk targets to a list if they (for example) have filtered the list down to what they want to add to the group. So: a checkbox in the target list table header that when you click it, selects all targets.

Echoing David here: unit tests would be great!

Fingel:
Is this now tested and ready to go? We should merge this before it sits too long.

xuiris:
@Fingel, since last time we have only:
- improved the message feedback after adding targets to group
- wrote unit tests

Persisting the filter and adding a select all has not been done.

PR #119, opened 2019-07-16
Adding settings for JPEGs into .tmpl

zemogle:


Fingel:
I think the reason there are conflict is because master was merged in manually. You should only branch/merge development (master will be merged into development).

zemogle:
This was done because `development` had lagged behind `master`, I think due to the observing portal v3 upgrade.

Fingel:
I did this manually in this commit: https://github.com/TOMToolkit/tom_base/commit/f064ac8c1d92a464b35e87f7db79f608586878ea 
Does that negate the need for this PR?
Also, this will break people upgrading their TOMs regardless. Code should not assume these settings exist, but should instead use try/except blocks and use sensible defaults if the settings do not exist.

zemogle:
That's perfect. Thanks.

zemogle:
I just noticed that you did this commit on `master` only. Could you pull it into `development`? thanks.

PR #126, opened 2019-07-30
Allow target extra fields to be hidden from the target detail view

joesingo:
This commit allows users to mark certain `EXTRA_FIELDS` as 'hidden' so that they are not displayed in the target detail view:

```
EXTRA_FIELDS = [
    {'name': 'extrafield', 'type': 'string', 'hidden': True}
]
```

Hidden fields can still be accessed in the usual way with `target.extra_fields`; the only difference is that they are excluded in the target detail template.

This is needed for the asteroid tracker application, where we are storing long-form descriptions of targets as an extra field for use outside of the TOM.

If merged, the [custom target fields documentation](https://github.com/TOMToolkit/tomtoolkit.github.io/blob/master/docs/target_fields.md) would need updating.


Fingel:
Another good one, thanks!

PR #137, opened 2019-08-07
Make extra field values in Target.extra_fields the correct type

joesingo:
Previously the `extra_fields` dictionary for the `Target` model had all
values as strings. E.g. `'False'` (as a string) instead of `False` for a
boolean field. This commit makes sure the types are correct.

(Note: The only place this dictionary was used in the `tom_base` code is
in the target detail template. The template still displays properly
after this change.)


Fingel:
Awesome, thank you.

PR #139, opened 2019-08-07
Fix/soar submission

dmcollom:
Addresses issues with submitting to SOAR due to interface changes and LCO data inconsistencies (SOAR instrument names aren't prepended with telescope class)
Improves instrument selection with friendlier choices
Adds SOAR site to visibility calculator

One concern with this is that for the visibility calculator, SOAR shows up as "LCO (Cerro Pachon)". Politically, that may not be ideal. The alternative, which may be more palatable, would be to create a SOAR facility module that inherits from LCO.

### DIFF
--- a/tom_observations/facilities/lco.py
+++ b/tom_observations/facilities/lco.py
@@ -67,6 +67,12 @@ SITES = {
         'latitude': 20.706,
         'longitude': -156.258,
         'elevation': 3065
+    },
+    'Cerro Pachón': {
+        'sitecode': 'sor',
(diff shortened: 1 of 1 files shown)


Fingel:
This might be better off as it's own SOAR module. Like you said politically it will look better, plus you can clean up some of the soar specific code branches.

rachel3834:
Having a SOAR module seems like a cleaner solution to me, and would make it easier for users to interpret the visibility results without needing to know whose telescopes are on which mountain.

PR #141, opened 2019-08-08
Use local defaulted settings instead of global settings module

Fingel:


tom_dataproducts/models.py
```diff
@@ -21,18 +17,28 @@
...
+except AttributeError:
+    THUMBNAIL_MAX_SIZE = (0, 0)
+
+try:
+    THUMBNAIL_DEFAULT_SIZE = settings.THUMBNAIL_DEFAULT_SIZE
+except AttributeError:
+    THUMBNAIL_DEFAULT_SIZE = (200, 200)
+
 
 def find_img_size(filename):
```
Fingel: @zemogle Does this method ever do anything? Even if you assume that the value is in settings.py, you would never reach the except block. Or do we need to run this except block if the setting doesn't exist?

zemogle: With this fix, me and my intern think it should catch all eventualities.

PR #143, opened 2019-08-08
Changed extension header for identifying valid fits files

dmcollom:
At present, thumbnails cannot be generated for raw images, only for LCO-reduced images. This PR does the following:

- Changes the header used to identify valid images to `EXTNAME` instead of `XTENSION`, which appears to be a BANZAI-generated header keyword.
- Uses magic to validate filetype instead of relying solely on astropy's fits library, which we do elsewhere.

It also addresses the Plotly update separating the properties for coastline and landmass appearances for the observation distribution plot.

### DIFF
--- a/tom_dataproducts/models.py
+++ b/tom_dataproducts/models.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 import os
 import tempfile
+import magic
 
 from astropy.io import fits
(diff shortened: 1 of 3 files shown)


dmcollom:
> In theory this is good but the tests should pass.

No kidding.

PR #145, opened 2019-08-12
Feature/ci linting

dmcollom:
I've never set up linting for TravisCI before, so hopefully this is the preferred way to do so.

Also included are fixes for the million files that didn't pass the style checks.


dmcollom:
I included it as a dependency so that contributors can easily run it locally, but I will instead make a note of what to install/run before PRing in the docs.

Fingel:
I'd rather leave it up to each contributor. There are many tools that do the job. For example, I use flake8. Pycodestyle is probably the most basic one.

dmcollom:
Well on that note, should we instead use a more full-featured linter? I chose pycodestyle because it's what Astropy uses, but I'm open to something a bit more robust.

Fingel:
I think pycodestyle is fine. Flake8 is the same thing with a little extra, not really a big deal.

PR #150, opened 2019-08-14
Separated base observation forms into imaging and spectroscopy, updat…

dmcollom:
…ed views to support different observation types. 

- Provides a decent proof of concept for creating custom observation forms for an existing facility.
- Only interface changes are requiring `observation_types` in a facility module, and adding a `get_form` abstract method.
- Imaging and Spectroscopy forms both inherit from facility's base form.
- Added hidden `observation_type` field to `GenericObservationForm` for determination of form selection.
- Renamed `observation_type` in `LCOObservationForm` to `observation_mode`. The field is denoted as `observation_type` in the API, but is `observation_mode` in the Observation Portal.
- Made NRES submissions minimally functional.

There is no accompanying change to the documentation, but I will do so immediately if this is an agreeable solution for dividing observation types into separate forms.


dmcollom:
I have a branch with the updates to the observation module documentation [here](https://github.com/TOMToolkit/tomtoolkit.github.io/pull/19), and I will merge it when we have our next release.

PR #154, opened 2019-08-21
Feature/alerce

dmcollom:
Did this to figure out the kinks in ALeRCE so that I'll be able to walk any hack session attendees through this as an exercise.

A note: the construction of the payload is so obnoxious because even submitting an empty dictionary for the `filters`, `coordinates`, or `dates` sections will result in a 500.


Fingel:
Good luck!

PR #163, opened 2019-09-09
Replace libmagic with ext guessing for fits and exception handling

Fingel:


Fingel:
Works for me, Travis seems to disagree with me on what tests pass anyway.

PR #165, opened 2019-09-09
Fix/data upload

dmcollom:
This PR addresses the following:

- Removes libmagic dependency, replaces with mimetypes
- Reads facility and date-obs from fits headers or CSV comments as applicable
- Drastically simplifies data product upload form
- Reorders tom_targets.tests dependencies for some reason

### DIFF
--- a/tom_dataproducts/data_processor.py
+++ b/tom_dataproducts/data_processor.py
@@ -37,9 +37,6 @@ class DataProcessor():
         """
 
         mimetype = mimetypes.guess_type(data_product.data.path)[0]
-        print(mimetype)
-        print(mimetypes.knownfiles)
-        print(mimetypes.read_mime_types(data_product.data.path))
         if mimetype in FITS_MIMETYPES:
             return self._process_spectrum_from_fits(data_product)
         elif mimetype in PLAINTEXT_MIMETYPES:
--- a/tom_dataproducts/data_processor.py
+++ b/tom_dataproducts/data_processor.py
@@ -37,6 +37,7 @@ class DataProcessor():
         """
 
         mimetype = mimetypes.guess_type(data_product.data.path)[0]
+        print(mimetype)
         if mimetype in FITS_MIMETYPES:
             return self._process_spectrum_from_fits(data_product)
         elif mimetype in PLAINTEXT_MIMETYPES:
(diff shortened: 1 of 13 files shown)


tom_setup/management/commands/tom_setup.py
```diff
@@ -143,6 +144,12 @@ def create_public_group(self):
         group.save()
         self.ok()
 
+    def add_fits_mimetypes(self):
```
Fingel: mimetypes.add_type is not persistent, you need to do it somewhere it will be executed every time.

tom_dataproducts/tests/tests.py
```diff
@@ -299,57 +278,56 @@ class TestDataProcessor(TestCase):
     def setUp(self):
         self.target = TargetFactory.create()
         self.data_product = DataProduct.objects.create(
-            target=self.target,
-            data=SimpleUploadedFile('afile.fits', b'somedata')
+            target=self.target
         )
         self.data_processor = DataProcessor()
+        self.test_file = SimpleUploadedFile('afile.fits', b'somedata')
+        mimetypes.add_type('image/fits', '.fits')
```
Fingel: This needs to be done in the application code, not the tests.

tom_dataproducts/data_processor.py
```diff
@@ -7,124 +8,142 @@
 from specutils import Spectrum1D
 import numpy as np
 
-from tom_observations.facility import get_service_class
+from tom_observations.facility import get_service_class, get_service_classes
 from .exceptions import InvalidFileFormatException
 
 
+FITS_MIMETYPES = ['image/fits', 'application/fits']
+PLAINTEXT_MIMETYPES = ['text/plain', 'text/csv']
```
Fingel: Instead of restricting filetypes to the extensions .txt and .csv you might be better off trying to read the table every time and just catching the exception if it fails. Your call though.

dmcollom: It's used for routing to different behaviors based on filetypes so I guess I'd go with both. I'll make the change.

tom_dataproducts/tests/tests.py
dmcollom: Ah, I put it here because I didn't realize it wasn't persistent. I'll move it.

tom_setup/management/commands/tom_setup.py
Fingel: This should still be removed.

Fingel: This should still be removed.

dmcollom: Whoops. Removed this.

PR #166, opened 2019-09-09
Feature/sphinx

dmcollom:
This adds readthedocs/Sphinx support. Does not include a step in the build process to check contributed documentation, but should in the future.

The built version of the docs can be found at tom-toolkit.readthedocs.io for a preview of the documentation. @Fingel as soon as you create a ReadTheDocs account I can add you as an admin.

All docs were ported over from the TOM Toolkit website, and the link to `Documentation` will need to be updated to point here.


Fingel:
This is awesome!
Two of the links in the navigation (about and contributing) still link to the old docs.
This is a huge improvement.

dmcollom:
Good catch on the contributing guide, I ported it.

I left the About on the website because I think there's still value in having the nice pretty website as a landing page in addition to the generic ReadTheDocs site--I'm certainly not married to that opinion, though, and it's easily moved if we choose to do so. I also left the Plugins page, given that that isn't official yet.

Fingel:
I think we should definitely only have 1 website. We already have on lco.global, I would say we decommission the Jekyll one as soon as this is ready.

dmcollom:
Okay, sounds good. I've finished moving everything over--there's a built and slightly cleaner version up on tom-toolkit.readthedocs.io if you want to confirm.

PR #172, opened 2019-09-10
Consolidated target creation

dmcollom:
Very quickly addressing Andy's concern about ambiguity in target creation. This consolidates the three methods of target creation (manual, import, and catalog search) into one button/dropdown.


Fingel:
This works, but I actually think drop down navigation menus are considered bad UX.

PR #178, opened 2019-09-14
Changed prompt casing to be consistent

kurczynski:


dmcollom:
I believe this was intentional by @cmccully, because N is the default and the casing is meant to denote that. Curtis, can you confirm?

cmccully:
The default is supposed to be caps.

Fingel:
This is a common practice for cli interfaces and is intentional.

PR #182, opened 2019-09-17
Feature/airmass sidereal

jfrostburke:
Changes sidereal airmass plotting to use astroplan instead of pyephem. In addition to adding code, significantly reorganizes `tom_observations/utils.py` to make non/sidereal distinction clearer.

Tested on local copy and functioning for sidereal and non-sidereal targets.

If y'all have comments on style (or if I messed up docstring formatting or something) please let me know, any feedback is appreciated.


Fingel:
This PR would be a lot clearer if it removed the dependency on pyephem completely. As of now it looks like it adds a dependency on astroplan, replaces some functionality, but the toolkit still requires pyephem for quite a few functions. `tom_observations/utils.py` is the main culprit here.
Any way we can have a single dependency on astroplan?

dmcollom:
Yet again, pyephem was only used in an attempt to support both sidereal and non-sidereal. Astroplan does not do both, and presumably won't for quite a while. I think I would support a larger refactor that puts visibility calculations into a templatetag in a separate repo or something. Or just non-sidereal visibility. Or we remove non-sidereal visibility altogether and let the 5% of the community that cares figure it out.

Ultimately I'm just a little sick of getting hassled about PyEphem, and want it out as quickly as possible so I never have to hear about it again.

Fingel:
Re: Pyephem. I feel your pain. Would it be feasible to pull out the stuff you used pyephem for and have it be a optional plugin for non-sidereal support? And then let the base project use astroplan?

jfrostburke:
Having a non-sidereal plugin seems reasonable to me (speaking as an extragalactic astronomer). I don't know what the plumbing for that would look like though -- obviously it's easy enough to get rid of the pyephem/non-sidereal code here and throw an exception if you try to plot the airmass for non-sidereal targets, but I don't know how to implement it correctly if the user does have the plugin. 

It does seem insane to me that there's not a better way to plot airmass/get ephemerides of non-sidereal objects. It looks like JPL Horizons has an [astroquery module](https://astroquery.readthedocs.io/en/latest/jplhorizons/jplhorizons.html) to get ephemerides, but it looks like it only works if the object is already in JPL Horizons. We should talk to a space rocks person to see what the right thing to do is.

dmcollom:
Yes, 100% in favor of some kind of `tom_extras`.

I think we could simply not show the visibility for non-sidereal targets by default. The plot only shows up on the `target_detail` template, and we have the target_type, so it'd be a simple change there. Have the plotting templatetag take a target_type and route it to the correct visibility calculation--in this case, the only one we have.

This could also probably be condensed into a single method, given that most of the complexity was previously centered around determining target type.

jfrostburke:
Just an update that I removed non-sidereal plotting and adding error checking to make sure it can't happen in tom_base: `target_plan` is removed by default in `target_detail.html` for nonsidereal targets, if they put the form in manually I invalidate it if `target.type != 'SIDEREAL'`, if they run it programmatically it logs an angry red comment telling them not to do this and returns a blank visibility dic.

I also made a separate repo for nonsidereal airmass plotting (currently at https://github.com/jfrostburke/tom_nonsidereal_airmass) that uses the old pyephem code. I still have some work to do (need to remove pyephem from the tests, put my nonsidereal repo on pypi, stylecheck this PR to appease the git gods), but am aiming to finish this Monday so it can be in 1.0.

jfrostburke:
Pyephem dependency removed after I updated the tests with help from David, merged with dev after the targetnames update, and also updated installation instructions in the nonsidereal repo to be accurate.

tom_observations/tests/tests.py
```diff
@@ -101,74 +102,52 @@ def test_update_individual_target_observations_for_facility(self):
...
-
-    def test_get_rise_set_invalid_params(self):
+        self.start = datetime(2018, 10, 9, 13, 56, 16)
+        self.interval = 10
+        self.airmass_limit = 10
+
+    def test_get_astroplan_sun_and_time(self):
+        end = self.start + timedelta(days=2)
+        sun, time_range = get_astroplan_sun_and_time(self.start, end, self.interval)
+        self.assertIsInstance(sun, SkyCoord)
```
dmcollom: I think these tests should also confirm that the sun time ranges are as expected.

tom_observations/utils.py
```diff
@@ -1,125 +1,25 @@
...
-
-    :param rise_sets: array of tuples representing set of rise/sets to search
-    :type rise_sets: array
-
-    :param time: time value used to find the next rise, in UNIX time
-    :type time: float
-
-    :returns: Soonest upcoming rise/set with respect to the given time
-    :rtype: tuple
+def get_visibility(target, start_time, end_time, interval, airmass_limit):
```
dmcollom: Wasn't this called get_sidereal_visibility at one point? If not, it probably should be.

dmcollom:
Looks good aside from the two little changes. I also went ahead and remove pyephem from the setup.

Is there a corresponding repo for the non-sidereal pyephem garbage I wrote, or do you want me to work on that?

tom_observations/tests/tests.py
jfrostburke: Updated in most recent push. The `time_range`s were too long to hardcode in their values nicely to check against, so I checked their length and then every nth value (every 20th value for the shorter one, every 50th value for the longer one).

tom_observations/utils.py
jfrostburke: Updated in most recent push throughout the code (tests and demo TOM website run).

jfrostburke:
Whoops thanks for catching that, I thought I did that when I added the `astroplan` dependency but I guess not. Both of your comments have been addressed.

The nonsidereal plotting now lives at https://github.com/jfrostburke/tom_nonsidereal_airmass.

tom_observations/tests/tests.py
dmcollom: So unless I'm mistaken on what you're trying to do, there's an `assertAlmostEqual` that takes a number of significant figures to equate, that might do a little better for your case. Props for being clever though.

dmcollom:
Awesome, thanks for doing that! Just a couple more things:

* Should that repo be under the TOMToolkit umbrella? I personally would think so.
* Is it published to PyPi?
* And the larger question, do we want that do be a more all-compassing repo that has a bunch of extras in the future, or should it solely contain non-sidereal visibility?

jfrostburke:
I agree it should be under the TOMToolkit umbrella. I was about to say "I'm not in the organization" but Austin just invited me. It's not currently on PyPI, although my initial setup mostly copied from `tom-antares` so it could be on PyPI without too much effort (I think, I've never published a package before).

I definitely support a `tom_extras` repo or something that collects all the add-ons, since we've already got a Gemini plugin, a Liverpool Telescope plugin, the Antares plugin, and now this.

jfrostburke:
Okay, made the nonsidereal airmass repo under the TOMToolkit organization (https://github.com/TOMToolkit/tom_nonsidereal_airmass) and added it to PyPI.

dmcollom:
Awesome, great work. This looks fine to me to merge.

PR #183, opened 2019-09-18
Fix typos

irasnyd:
I am working with Edward Gomez to get his TOM Toolkit project running on Kubernetes. While reading the documentation, I found tons of typoes. Here are fixes for all of the ones I spotted.

### DIFF
--- a/docs/advanced/backgroundtasks.md
+++ b/docs/advanced/backgroundtasks.md
@@ -7,12 +7,12 @@ you click a button or link in the TOM your browser constructs a web request, whi
 is then sent to the webserver running your TOM. The TOM receives this request and
 then runs a bunch of code, ultimately to generate a response that gets sent back
 to the browser. This response is what you see when the next page loads. For the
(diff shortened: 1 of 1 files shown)


docs/advanced/backgroundtasks.md
```diff
@@ -79,7 +79,7 @@ You can now run the server with:
     ./src/redis-server
 
 
-If you are running linux, most likely Redis is included with your distribution via
+If you are running Linux, most likely Redis is included with your distribution via
 it's package manager. For example:
```
dmcollom: I know you didn't add this line, but this should be its rather than it's if you wanted to do that now.

dmcollom:
Fixed your branch name. :stuck_out_tongue_winking_eye: Thanks for fixing these!

docs/advanced/backgroundtasks.md
irasnyd: I updated the commit with this change, and another that I found that needed the same fix. Thanks.

PR #185, opened 2019-09-19
Fix/snex feedback

dmcollom:
Added a small subset of the changes discussed in the response to Andy's document. Among those:

- Headers on pages indicating their purpose
- Hints
- Required email
- Changed some verbiage
- Reordered some import statements for reasons

### DIFF
--- a/tom_base/settings.py
+++ b/tom_base/settings.py
@@ -226,6 +226,9 @@ THUMBNAIL_MAX_SIZE = (0, 0)
 
 THUMBNAIL_DEFAULT_SIZE = (200, 200)
 
+HINTS_ENABLED = False
+HINT_LEVEL = HINTS_ENABLED * 20
+
 try:
     from local_settings import *  # noqa
 except ImportError:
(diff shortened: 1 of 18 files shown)


tom_base/settings.py
```diff
@@ -226,6 +226,9 @@
 
 THUMBNAIL_DEFAULT_SIZE = (200, 200)
 
+HINTS_ENABLED = False
+HINT_LEVEL = HINTS_ENABLED * 20
```
Fingel: Well, TIL that `True * 20 == 20` and `False * 20 == 0`. Clever, probably too clever. This is the kind of thing that breaks with a major Python release. 
Much simpler would be to simply write your own `add_message` function that checks settings.HINTS_ENABLED.

dmcollom: I actually spent quite a bit of time trying to override the MessageMiddleware to intercept and remove hint messages, but that proved to be more trouble than it was worth. This was the five minute solution. You're right, something in the middle is probably the better option...I'll update it.

Fingel: Can you do something like this?

```python
import messages
def add_hint(request, message, level=settings.HINT_LEVEL, *args, **kwargs):
    if settings.HINTS_ENABLED:
        messages.add_message(request, level, message, *args, **kwargs)
````

dmcollom: Yeah, that would have been the obvious, sane choice.

dmcollom: Fixed this, and added corresponding documentation for the setting.

PR #193, opened 2019-09-27
Thumbnail bug-fix for non-local file system

zemogle:
There is a bug when files are non-locallly handled (i.e. via S3). Currently filenames and paths are passed. This PR passes file objects to the jpeg creation service.

This requires `fits2image==0.4.1` which is at present also a pending PR https://github.com/LCOGT/fits2image/pull/4


dmcollom:
Looks fine--can you fix the tests though?

zemogle:
Good point. I've fixed them up now.

PR #201, opened 2019-10-04
add TNS broker

griffin-h:
I made a first attempt at an alert broker module to query the [Transient Name Server](https://wis-tns.weizmann.ac.il) by name, coordinates, and/or discovery date.


Fingel:
Excellent! Thank you Griffon.

PR #207, opened 2019-10-18
Fix/dataproduct tags

dmcollom:
This pull request contains the following:

* Renames DataProduct.tag -> DataProduct.data_product_type to avoid confusion with Target.tag. This may or may not be ill-advised.
* Adds DATA_PRODUCT_TYPES to settings.py and removes constraints on model field.
* Adds migration script to rename column and ensures consistency in type casing.


dmcollom:
Now that I think about it again, because I've agonized over this, data_product_type may not be the best decision either, as it implies a distinct value, and there's definitely overlap between some data products and some types.

dmcollom:
No, you aren't missing anything. Previously the data processing was reliant on the built-in types that were set up in the models themselves. Now, since we're making them user-configurable, yes, there is an expectation that either they keep the built-in ones (hence the warning comment in `settings.py`), or override the data processor.

I am by no means tied to the solution. It's very possible that I'm a little myopic on this as I tried to adapt the existing solution to fit Edward's desires. Happy to (and would prefer) have a larger discussion on this, since we were discussing redoing a lot of this code anyway.

Fingel:
Maybe instead of the `data_processor.py` class handing every data type we map data processors directly to types.  If someone wants to customize how photometry is processed they just change the setting to point to a different class for photometry. Or if you they have a completely different type they can add it to the settings and map it to a processor class.
I think the entrypoint is just one level too high currently, if that makes sense.

dmcollom:
I think that makes sense, and was very much what I was trying to get at with the original solution, but this is a better architecture.

dmcollom:
This now has the improvements that we're (hopefully) looking for. All data processors have to inherit from `DataProcessor` and implement `process_data`, which is what `run_data_processor` calls.

@Fingel let me know if you think there are any shortcomings here, and if the docs adequately explain it. I'll update docstrings in the meantime, since those are yet to be done.

PR #211, opened 2019-10-23
More options on data upload form

griffin-h:
Uploading reduced data products currently requires that they be in a very specific format, which is not standard across different facilities. I added some more options to the data upload form that allows the user more flexibility in the file format. All these formats are already available in Astropy, so it was mostly a matter of piping them through to the form. This certainly doesn't cover all possible formats, but it's an improvement over having to reformat every file I want to upload to match the TOM Toolkit format.

All plain text files:
 - table format
 - time format
 - where to find the time (column header or header keyword)

Photometry:
 - column headers for magnitude, error, and filter

Spectroscopy:
 - manually input date-obs
 - column headers for wavelength and flux (for plain text files)

Ideally, it would be nice to dynamically update the form fields based on the type of data being uploaded (tag), but I don't know how to do this.

This may also make the "FITS facility" framework obsolete, but I left that alone in case it's needed for something else.

### DIFF
--- a/tom_dataproducts/data_processor.py
+++ b/tom_dataproducts/data_processor.py
@@ -28,7 +28,7 @@ mimetypes.add_type('application/fits', '.fits')
 mimetypes.add_type('application/fits', '.fz')
 
 
-def run_data_processor(dp):
+def run_data_processor(dp, form_data):
     try:
         processor_class = settings.DATA_PROCESSOR_CLASS
     except Exception:
@@ -43,7 +43,7 @@ def run_data_processor(dp):
     data_processor = clazz()
 
     if dp.tag == SPECTROSCOPY[0]:
-        spectrum, obs_date = data_processor.process_spectroscopy(dp)
+        spectrum, obs_date = data_processor.process_spectroscopy(dp, form_data)
         serialized_spectrum = SpectrumSerializer().serialize(spectrum)
         ReducedDatum.objects.create(
             target=dp.target,
@@ -53,7 +53,7 @@ def run_data_processor(dp):
             value=serialized_spectrum
         )
     elif dp.tag == PHOTOMETRY[0]:
-        photometry = data_processor.process_photometry(dp)
+        photometry = data_processor.process_photometry(dp, form_data)
         for time, photometry_datum in photometry.items():
             for datum in photometry_datum:
                 ReducedDatum.objects.create(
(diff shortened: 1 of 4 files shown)


dmcollom:
Hey Griffin, really appreciate you taking a look at this. This is a good reminder that we need to improve this part of the UI/UX, as well as dealing with incoming values more elegantly.

First off, I'm working on a refactor of the architecture of the processing code that's going to wreak havoc on your fork. I went ahead and did the merge locally, and with your permission I'll push it to this branch.

So my main concern here is that the form has a few too many options, many of which are only used selectively. I was already concerned when I made a form that had facility and observation date, because those weren't applicable for our suggested photometry format. I guess the question for you is, how can we make this more flexible, because I assume some groups will want a one-click upload, and other won't. How do you think we should address this?

griffin-h:
Yes of course, please use this however you see fit (or not at all). I mostly wrote it for myself so I can import all my existing data, but I thought it might be useful for others as well.

To answer your question, I think the best way to reduce the number of fields is to have only the relevant fields appear when you click on one of the tags. This is what I was referring to above, but I don't know how to do this in Django or if it's even possible. So for photometry, you would have rows 1 and 2 appear; for spectroscopy, rows 1 and 3; and for the other tags (which are what I consider "raw data") there are no options.

I tried to give sensible defaults to each field, but maybe it would be a good idea to make those defaults configurable in the settings.py. That way, if one group has a standard format that they use, they could essentially have a one-click upload, because they would never need to change the fields from their defaults.

If you really want to reduce the number of fields to 0, the best way might be to loop through a number of guesses until one of them works. I can try to supply what the guesses should be for each of those fields, but the plain text formats people use are so non-standard, I have no idea what fraction of the time this would work.

Fingel:
> If you really want to reduce the number of fields to 0, the best way might be to loop through a number of guesses until one of them works. I can try to supply what the guesses should be for each of those fields, but the plain text formats people use are so non-standard, I have no idea what fraction of the time this would work.

This is essentially what I was suggesting during the post TOM Toolkit Workshop meeting a few weeks ago (@dmcollom you were remote). Essentially we write classes that take in a file and either can or cannot read them and return that back up. The framework will use the first class that successfully claims it can read a file. I imagined these would be per-facility or instrument but they don't have to be.

dmcollom:
Just to give you an update, because I have done some work on this, I have a branch that Austin is currently reviewing, that I hope will offer some measure of simplification. I've also already completed work allowing for customization of processing on a type-by-type basis. If you're curious to review it, the merged PR is #207 and the in-progress-in-my-spare-time branch is fix/dataproduct_upload. I'm sure they aren't perfect, so feedback is very welcome!

griffin-h:
@jchate6

PR #217, opened 2019-11-06
Updated templates to fix tom_setup

dmcollom:
A couple of changes in development introduced breaking changes for tom_setup. This corrects those, and proactively updates the access method to the TNS API key to be consistent.


jfrostburke:
Tested by installing an out-of-the-box TOM using this branch, and it successfully fixes previous issues which occurred on the initial `./manage.py migrate`.

PR #223, opened 2019-11-19
New broker interface for the Gaia alerts system

rachel3834:
Interface to harvest alerts from the Gaia Alerts service:
http://gsaweb.ast.cam.ac.uk/alerts/alertsindex
packaged in the form of a TOM broker interface accessible through the browser-based UI, and incorporating the functionality developed by Lukasz Wyrzykowski's as a code hook.


dmcollom:
This looks great! I just have a few things:

1. Looks like it isn't passing style checks! It's just a few things, shouldn't be hard to fix.
2. You should definitely add the broker to `TOM_ALERT_CLASSES` in `alerts.py` so that it shows up by default in new TOMs.
3. You may consider renaming things to remove "alerts", such as `gaiaalerts.py` -> `gaia.py` and GaiaAlertBroker -> GaiaBroker. I understand that it's the official name, but we don't call them `MARSAlertBroker` or anything else, so matching consistency would be good.
4. This is definitely a personal decision, but it looks like a user can just submit nothing in the query params and get everything back (or rather, the first 20 results). You could, I believe, implement a `clean` on your query form that would allow you to do cross-field validation, if that were something you wanted.

rachel3834:
Thanks for the feedback!  I was wondering about how best to do that validation.  I think I've addressed all your points.

PR #238, opened 2020-01-24
Minor changes on ALeRCE Broker 

cvalenzu:
Cheking the ALeRCE broker plugin it was a bit outdated in the class taxonomy and with some queries, so I added some new features:

- Uses an API request to get the latest taxonomy for each classifier.
- Added relative date in hours instead of a fixed date (useful for early SN discovery)
- Sort the results by number of detections or probabilities.
- Give a maximum amount of retrieved pages to have a faster response time. 

image

It was passing all the code styles rules, don't know if the travis tests are passing.

Cheers~
Camilo.

### DIFF
--- a/tom_alerts/brokers/alerce.py
+++ b/tom_alerts/brokers/alerce.py
@@ -18,7 +18,7 @@ SORT_CHOICES = [("nobs", "Number Of Epochs"),
                 ("pclassearly", "Early Probability")]
 
 PAGES_CHOICES = [
-    (i, i) for i in [5, 1, 10, 15]
+    (i, i) for i in [1, 5, 10, 15]
 ]
 
 RECORDS_CHOICES = [
--- a/tom_alerts/brokers/alerce.py
+++ b/tom_alerts/brokers/alerce.py
@@ -3,25 +3,43 @@ import requests
 from django import forms
 from crispy_forms.layout import Layout, Div, Fieldset
 from astropy.time import Time, TimezoneInfo
+import datetime
 
 from tom_alerts.alerts import GenericQueryForm, GenericBroker, GenericAlert
 from tom_targets.models import Target
 
-ALERCE_URL = 'http://alerce.online'
-ALERCE_SEARCH_URL = 'http://ztf.alerce.online/query'
+ALERCE_URL = 'https://alerce.online'
+ALERCE_SEARCH_URL = 'https://ztf.alerce.online/query'
+ALERCE_CLASSES_URL = 'https://ztf.alerce.online/get_current_classes'
 
-RF_CLASSIFIERS = [(None, "None")] + [(k, k) for k in ["CEPH", "DSCT", "EB", "LPV", "RRL", "SNe", "Other"]]
-STAMP_CLASSIFIERS = [(None, "None")] + [(k, k) for k in ["AGN", "SN", "VS", "asteroid", "bogus"]]
+SORT_CHOICES = [("nobs", "Number Of Epochs"),
+                ("lastmjd", "Last Detection"),
+                ("pclassrf", "Late Probability"),
+                ("pclassearly", "Early Probability")]
+
+PAGES_CHOICES = [
+    (i, i) for i in [5, 1, 10, 15]
+]
+
+RECORDS_CHOICES = [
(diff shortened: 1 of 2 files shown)


tom_alerts/brokers/alerce.py
```diff
@@ -43,31 +61,69 @@ class ALeRCEQueryForm(GenericQueryForm):
...
+        required=False,
+        label='Relative date of object discovery.',
+        widget=forms.TextInput(attrs={'placeholder': 'Hours'})
+    )
+    sort_by = forms.ChoiceField(
+            choices=SORT_CHOICES,
+            required=False,
+            label='Sort By'
+    )
+    max_pages = forms.TypedChoiceField(
```
dmcollom: This is reeeeeally nitpicky, but if you add `initial=5` to the TypedChoiceField, it will default to 5 and the list will be in order.

tom_alerts/brokers/alerce.py
```diff
@@ -213,6 +348,6 @@ def to_generic_alert(self, alert):
             name=alert['oid'],
             ra=alert['meanra'],
```
dmcollom: I realize you didn't touch this and this was my code, but I'm curious about the meanra/meandec items. It looks like some 2018 objects come back with meanra/meandec, but the ZTF19 objects don't have that data. Is that an API change? While it isn't 100% necessary, it'd be nice to be able to populate the results table if we can.

tom_alerts/brokers/alerce.py
cvalenzu: The max of 5 pages its more a practical thing, there are queries you can make that will return thousands pages and will take a lot of time.

Sorting them out and giving the `records_per_page*max_pages` as the maximum records its a better strategy to retrieve useful ztf objects.

tom_alerts/brokers/alerce.py
cvalenzu: The `meanra `and `meandec` are the aggregated `ra` and `dec` for an object, there must be in every record on the API, if there those fields aren't filled maybe is more a bug on our pipeline rather than a problem on the API.

If you have some examples would be really helpful if you can share them with us.

tom_alerts/brokers/alerce.py
dmcollom: Sorry, what I mean to say is, the list of choices for the TypedChoiceField, as defined on line 21, is ordered [5, 1, 10, 15], presumably to make 5 the default, which I think is just fine. This order is also reflected in the dropdown on the UI.

If you add the `initial` kwarg to the TypedChoiceField, you can put them in numerical order while still showing 5 as the initial value.

Completely on board with 5 pages being the default max--it's a very practical decision.

tom_alerts/brokers/alerce.py
dmcollom: Of course! So the test query I ran when I pulled down your branch was simply Early Classifier SN, Probability 0.9, 20 records per page, 5 pages.

On the first page, there are five ZTF18 candidates, and 15 ZTF19 candidates. 3/5 of the ZTF18 candidates has a meanra and meandec, and none of the ZTF19 candidates have either.

One of the ZTF18 candidates with meanra/meandec is ZTF18aauselm.
One of the ZTF18 candidates without meanra/meandec is ZTF18aclbfow.
One of the ZTF19 candidates without meanra/meandec is ZTF19acekhzt.

I briefly looked at the response to see if the TOM module I wrote just isn't processing them properly, but I figured I'd check with you about a possible API change in the past that I missed before I went digging.

tom_alerts/brokers/alerce.py
cvalenzu: I've pushed a new commit changing that, was doing some tests and forgot to change them.

tom_alerts/brokers/alerce.py
cvalenzu: We check those objects and was a problem with the pipeline, its not an issue with the TOM or the API.
We will try to fix it ASAP, thanks for finding the bug we missed them.

dmcollom: Awesome! This all looks great, I'll get it merged and it'll be in the next release.

PR #250, opened 2020-02-13
Updated formula for cone search, modified everything to use radians

dmcollom:


dmcollom:
Fixed the tests and made them more applicable, also updated the search itself to be preceded by a square search to improve performance.

PR #256, opened 2020-02-25
Feature/row level permissions

dmcollom:
This pull request adds row-level permissions provided the pertinent setting is toggled, and bypasses such logic otherwise.

By default, the only models restricted by group permissions are targets. This PR adds a second mode, which restricts models by group permissions for observations, data product, and reduced datums.


dmcollom:
Thank you, I pretty much just put the first thing that came into my head just to move forward and never revisited it. I will update it.

Issue #252, opened 2020-02-20
Astropy 4.0 required


mpgh:
During the installation on a system with an older astropy version, some packages require astropy 4.0 which could be added to the requirements.txt

More specifically the following version is incompatible:

ERROR: gwcs 0.12.0 has requirement astropy>=4.0, but you'll have astropy 3.2.2 which is incompatible.
ERROR: specutils 0.7 has requirement astropy>=4.0, but you'll have astropy 3.2.2 which is incompatible.

**Desktop (please complete the following information):**
 - OS: Linux
 - Python Version 3.7.4

dmcollom:
This has been addressed in #258 and is in release 1.3.1. Thanks for bringing this to our attention, @mpgh!

PR #258, opened 2020-02-28
Pinning astropy and specutils versions to stated compatible versions

dmcollom:
When starting the TOM Toolkit with a version of Astropy < 4.0, astropy warns the user that specutils 0.7 is incompatible. We also saw a brief issue in spectroscopy uploads with Astropy 4.0 that required a fix, so it's safer just to pin both right now.

This also addresses #252


dmcollom:
I would think that if we're pinning a bunch of stuff we should thoroughly test it each time, and that seems out of scope for this two-line fix that addresses a specific issue.

But yeah, at some point we should.

PR #260, opened 2020-03-04
Added LT stub and made minor modification to observation type tabs to…

dmcollom:
… ensure they aren't displayed


dmcollom:
It's intentional--ideally the default set of classes should include everything, so new users are aware of everything out of the gate (in my opinion). Not sure why it was omitted in the first place, but I'm taking the opportunity to add it now.

PR #261, opened 2020-03-05
Added Jamie's moon distance plot

dmcollom:
Adapted the plot from snex2 for lunar distance and added it to the TOM Toolkit. Also made the code less Ascension and more Blue Train.

### DIFF
--- a/tom_observations/facilities/lt.py
+++ b/tom_observations/facilities/lt.py
@@ -1,6 +1,6 @@
 from crispy_forms.layout import Layout, HTML
 
-from tom_observations.facility import GenericObservationForm
+from tom_observations.facility import GenericObservationFacility, GenericObservationForm
 
 
 class LTQueryForm(GenericObservationForm):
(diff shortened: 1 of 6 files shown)


jfrostburke:
Added code looks good, however when I installed this branch to test manually on an out-of-the-box TOM I ran into two errors, both of which stem from an earlier [commit](https://github.com/TOMToolkit/tom_base/commit/75df05e8f5b95aec95ab03c84435ab8b12c7d031):
1. `manage.py migrate` did not work because of the `settings.py` file: 
`  File "/home/jburke/TOM/tom_moon_test/tom_moon_test/settings.py", line 240
    'tom_observations.facilities.soar.SOARFacility`
2. After fixing this the TOM started normally. I created a target, and its page displayed correctly (including the moon plot). However when I tried to generate an airmass plot I got the error `'LTFacility' object has no attribute 'get_observing_sites'`.

Issue #218, opened 2019-11-08
Bootstrap Error Messages


jfrostburke:
**Related Problem**
The TNS harvester, which is included by default in the toolkit, requires an API key, which is not included by default. Currently if you try to import a target from the TNS without setting up the API key, then you get a Django error page.

**Suggested Improvement**
It would be friendlier if, instead of crashing with a Django error page, a Bootstrap error message popped up that said, "You can't use this harvester unless you add a TNS API key to your TOM" or something like that.

**Additional Request**
It would also be great to have some documentation about the correct way to add Bootstrap error messages to your TOM more generally. As an example also related to the TNS harvester, currently if you try to import a supernova using a name that doesn't exist in the TNS database (e.g. '3019np', or 'SN2019np' instead of just '2019np') then it also crashes and gives you a Django error page, whereas a Bootstrap error message would be more helpful. Knowing how to add Bootstrap error messages which are accessible in production and out of debug mode would provide a better user experience in general.

PR #270, opened 2020-04-15
Fix/tns harvester message

dmcollom:
This addresses #218 finally. Makes the api_key acquisition a little more bulletproof and throws the proper exception to ensure more attractive exception handling.

Also added a corresponding task to document the use of the ImproperCredentialsException.

@jfrostburke, can you test the actual functionality of the harvester? I just tested the errors--I don't have a TNS API key.

### DIFF
--- a/tom_catalogs/harvesters/tns.py
+++ b/tom_catalogs/harvesters/tns.py
@@ -7,29 +7,39 @@ from collections import OrderedDict
 from django.conf import settings
 
 from tom_catalogs.harvester import AbstractHarvester
+from tom_common.exceptions import ImproperCredentialsException
+
+TNS_URL = 'https://wis-tns.weizmann.ac.il'
+
+try:
+    TNS_CREDENTIALS = settings.BROKER_CREDENTIALS['TNS']
+except (AttributeError, KeyError):
+    TNS_CREDENTIALS = {
+        'api_key': ''
+    }
 
 
 def get(term):
-    api_key = settings.BROKER_CREDENTIALS['TNS_APIKEY']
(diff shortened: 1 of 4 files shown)


tom_catalogs/harvesters/tns.py
```diff
@@ -8,29 +7,38 @@
...
-        # construct the list of (key,value) pairs
-        get_data = [('api_key', (None, api_key)),
-                    ('data', (None, json.dumps(json_file)))]
+    response = requests.post(get_url, files=get_data)
+    response_data = json.loads(response.text)
 
-        response = requests.post(get_url, files=get_data)
-        response = json.loads(response.text)['data']['reply']
-        return response
+    if 400 <= response_data.get('id_code') <= 403:
```
dmcollom, line 38: While I would've loved to use response.status_code, the TNS instead chooses to return a 200 and puts the actual 401 in the response text, which is why this looks like this.

jfrostburke:
Response with API key:
Works as expected, correctly imports targets.

Response without API key:
This does show the Django bootstrap error message as expected, but the message isn't very helpful for troubleshooting. Currently it tells me
```
There was a problem authenticating with TNS: Unauthorized. Please check you have the correct credentials entered into your FACILITIES setting. https://tomtoolkit.github.io/docs/customsettings#facilities
```

Two things about this:
1. The link is unhelpful and just redirects to https://tom-toolkit.readthedocs.io/en/stable/
2. The key needs to be entered into `settings.BROKER_CREDENTIALS['TNS']['api_key']`, not `settings.FACILITIES`. I think a new custom exception/error message should be defined for this.

jfrostburke:
Updated error message looks good, merging.

Issue #272, opened 2020-04-16
Admin should not be able to delete AnonymousUser


TimotheeGr:
I recently deleted by mistake the AnonymousUser then I could not see any page without being logged in.

It was quickly solved by a `./manage.py migrate` but still it would be better if AnonymousUser was hidden from the Users page I suppose.
(Or maybe it has a use I don't know about...)

dmcollom:
Hi @TimotheeGr,

Appreciate you bringing this up! I think we can all agree that it's a little confusing to arbitrarily have a user that wasn't explicitly created. I'm adding a description of what `AnonymousUser` is to the FAQ, along with a sentence describing the implications of removing it. It'll be in the documentation after the next release.

The AnonymousUser is created by our AuthN/AuthZ library (django-guardian) to allow unauthenticated users to view public pages--which should explain why deleting AnonymousUser caused the problem that it did for you.

Now, to your point--while the user was visible from the page, the list of users is only visible to a PI (who is also a Django admin), so I don't think it's an inherent problem. If the user were hidden away and only visible in the Django admin interface, then there's a security risk if the PI is unaware of the AnonymousUser. I think the lessons here are that with great power comes great responsibility, and that we need to update our documentation, but that the code should probably stay as is.

Issue #271, opened 2020-04-15
Every broker should show a link to documentation


TimotheeGr:
I am using several of the brokers available and I have to search for the name of each broker on internet to know more about it.

It would be easier if every broker had a link toward the webpage of the broker or some documentation, like what the MARS broker does:
![Capture d’écran de 2020-04-15 18-07-44](https://user-images.githubusercontent.com/51410054/79393928-4f1c9b80-7f44-11ea-9730-4fd828a634e2.png)

jchate6:
I believe this is done now, but will add it to the Broker Audit (Issue #640 ) as a thing to check for.

PR #273, opened 2020-04-17
Feature/more docs

dmcollom:
In addition to addressing #271 and #272, this adds a blurb about writing tests to the FAQ doc.


dmcollom:
> Tested with a TNS API key and targets import correctly.

Did you test the harvester or the broker? The change was to the broker.

jfrostburke:
Ah good catch, I was testing the harvester. When I test the broker it does not work, I get a `403 Client Error: Forbidden for url: https://wis-tns.weizmann.ac.il/api/get/api/get/search`. I get the same error regardless of what field I query on -- object name, cone search, or discovery date. I'll send you the API key so it's easier for you to test.

dmcollom:
@phycodurus Since Austin fixed the TNS broker and I just merged in `dev`, this should be ready to merge.

PR #284, opened 2020-06-09
Feature/doc refactor

dmcollom:


dmcollom:
@phycodurus FYI: I'm really only look for a review on the README-dev.md from you. Everything else would be pretty burdensome to review.

PR #327, opened 2020-09-02
Added script for setting default values for new TargetExtras

dmcollom:


phycodurus (approved):
the only thing I might add is an example invocation in the doc string

PR #364, opened 2020-10-23
bugfix/fix-key-issue-tns-alerts

blancmatter:
Fixes dictionary key issues in the TNS API key and also in the response from the TNS. Fully tested.

Performed as per the Contributing instructions git workflow;
Forked, Branched, Fixed, Committed, Merged, Pushed to Fork, Creation of pull request.

Hope this is all good!


tom_alerts/brokers/tns.py
```diff
@@ -91,7 +91,7 @@ def fetch_alerts(cls, parameters):
         else:
             public_timestamp = ''
         data = {
-            'api_key': settings.BROKER_CREDENTIALS['TNS_APIKEY'],
+            'api_key': settings.ALERT_CREDENTIALS['TNS']['api_key'],
```
dmcollom, line 94: So, the move is actually going to be to call this settings `BROKERS`, as currently our facility info is stored in `settings.FACILITIES` (it holds more than just credentials). Can you make the name `BROKERS` instead?

dmcollom (approved):

dmcollom (approved):

Issue #374, opened 2020-11-03
Lasair broker filter parameters appear to be required but indicate they are optional


rachel3834:
**Describe the bug**
When I tried to create a filter for the Lasair broker, the form indicates that only the query name and the object name are required parameters.   However, when I run the filter without coordinates, it produces an 'NoneType' object is not an iterator error.  If I add coordinates and a radius, I get the error below - perhaps because I didn't include any JSON in the form submission?

**To Reproduce**
1. On an out-of-the-box TOM, go to Alerts -> Lasair
2. Complete the form fields which indicate the values are required, but not the remaining fields
3. Click save, then try to run the filter. 

**Expected behavior**
If the search coordinates, radius and JSON fields are required, then these parameters should have stars indicating that this is the case.  

**Desktop (please complete the following information):**
- OS: Mac OSx Catalina 10.15.7
- Browser Chrome

**Additional context**
Diagnostic output below:

Request Method: | GET
-- | --
http://127.0.0.1:8000/alerts/query/2/run/
3.1.1
JSONDecodeError
Expecting value: line 2 column 1 (char 1)
/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py, line 355, in raw_decode
/Users/rstreet1/software/tom_venv/bin/python
3.7.4
[...]

PR #380, opened 2020-11-13
Added form validation to Lasair and added basic tests

dmcollom:
Addresses #374 .


rachel3834 (approved):
These tests cover the use-case that I'm immediately concerned about.  
Something to bear in mind is that the form itself could use some better explanations about what to enter into each field.  For example the Object Cone Search field should indicate the units formatting of the RA, Dec (sexigesimal and/or decimal degrees), and should say what search radius is applied.  Also the Name field is ambiguous, so it would be good to rename this as Target Name.

PR #404, opened 2021-01-07
Feature/add muscat support

eheinrich:
Add support for observing on LCO's Muscat imaging instrument.

Muscat observation requests require a few different fields than normal imaging requests. There are separate fields for exposure times, one for each channel of the instrument instead of a single exposure time, and instead of a filter optical element, a diffuser optical element must be specified for each channel. In addition, an exposure mode and a guide mode are also selectable.

I hope I've added enough comments into the observing form itself, but let me know if anything does not make sense. I also tried to keep the style consistent with the rest of the page, but of course let me know if anything looks off.

### DIFF
--- a/tom_observations/facilities/lco.py
+++ b/tom_observations/facilities/lco.py
@@ -472,9 +472,10 @@ class LCOMuscatImagingObservationForm(LCOBaseObservationForm):
 
     @staticmethod
     def instrument_choices():
-        return sorted(set([
-            (k, v['name']) for k, v in LCOMuscatImagingObservationForm._get_muscat_instrument().items()]),
-            key=lambda inst: inst[1])
+        return sorted([
+            (k, v['name']) for k, v in LCOMuscatImagingObservationForm._get_muscat_instrument().items()],
+            key=lambda inst: inst[1]
+        )
 
     @staticmethod
     def mode_choices(mode_type):
--- a/tom_observations/facilities/lco.py
+++ b/tom_observations/facilities/lco.py
@@ -394,7 +394,10 @@ class LCOImagingObservationForm(LCOBaseObservationForm):
     @staticmethod
     def instrument_choices():
         return sorted(
-            [(k, v['name']) for k, v in LCOImagingObservationForm._get_instruments().items() if 'IMAGE' in v['type']],
+            [
+                (k, v['name']) for k, v in LCOImagingObservationForm._get_instruments().items()
+                if 'IMAGE' in v['type'] and 'MUSCAT' not in k
+            ],
             key=lambda inst: inst[1]
         )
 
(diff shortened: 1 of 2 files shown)


tom_observations/facilities/lco.py
```diff
@@ -406,6 +410,122 @@ def filter_choices():
...
+                Div(
+                    'instrument_type', 'guider_mode', 'exposure_mode', 'exposure_count', 'exposure_time_g',
+                    'exposure_time_r', 'exposure_time_i', 'exposure_time_z', 'diffuser_g_position',
+                    'diffuser_r_position', 'diffuser_i_position', 'diffuser_z_position',
+                    css_class='col'
+                ),
+                css_class='form-row',
+            ),
+            Div(
+                HTML(f'''<br/><p>Static cadence parameters. Leave blank if no cadencing is desired.
```
eheinrich, line 452: I copied this HTML block from the `LCOBaseObservationForm` layout. Since this same text both here and in the other form (the `"Static cadence parameters. Leave blank if no cadencing is desired."` string), I was hoping to merge it in with the `static_cadencing_help` string, but wanted to ask first - why are these two separate strings?

tom_observations/facilities/lco.py
```diff
@@ -406,6 +410,122 @@ def filter_choices():
...
+    exposure_time_z = forms.FloatField(min_value=0)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # `filter` and `exposure_time` are in the parent form but the Muscat form does not need them
+        self.fields.pop('filter', None)
+        self.fields.pop('exposure_time', None)
+        self.fields['guider_mode'] = forms.ChoiceField(choices=self.mode_choices('guiding'))
+        self.fields['exposure_mode'] = forms.ChoiceField(choices=self.mode_choices('exposure'))
+        self.fields['diffuser_g_position'] = forms.ChoiceField(choices=self.diffuser_position_choices(channel='g'))
```
dmcollom, line 430: I'm only seeing "Out of Beam" as an option, which makes sense considering it's the only one available in the API response. Is this something that's going to change in the future?

dmcollom (approved):
Looks good! Improvement in the layout might be nice, but this looks great to me.

tom_observations/facilities/lco.py
eheinrich: There is also an "In Beam" option - this option is available in the dev observation portal `/api/instruments/` endpoint. I don't know why it's not available in prod at the moment, but I assume it will be at that some point. When it is available, then it will show up as an option in the `/api/instruments/` response for the diffuser optical elements, and will in turn then show up as an option in this dropdown.

tom_observations/facilities/lco.py
```diff
@@ -406,6 +410,122 @@ def filter_choices():
...
+
+    def layout(self):
+        return Div(
+            Div(
+                Div(
+                    'name', 'proposal', 'ipp_value', 'observation_mode', 'start', 'end', 'max_airmass',
+                    'min_lunar_distance',
+                    css_class='col'
+                ),
+                Div(
```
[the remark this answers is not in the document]
eheinrich: Thanks for the suggestion! Will do some tweaking.

tom_observations/facilities/lco.py
eheinrich: Hmm ok. I ended up just merging the two strings, and the resulting paragraph on the page looks the same. I'll just go ahead and do that, but let me know if you have any objections.

PR #413, opened 2021-02-02
update TNS URL

griffin-h:
The Transient Name Server recently migrated their server to AWS, and now they have a new URL. See https://www.wis-tns.org/content/tns-newsfeed#comment-wrapper-20280 for more details.

I updated all instances of the wis-tns.weizmann.ac.il URL to www.wis-tns.org throughout the package.

### DIFF
--- a/tom_alerts/brokers/tns.py
+++ b/tom_alerts/brokers/tns.py
@@ -7,8 +7,9 @@ from datetime import datetime, timedelta
 from crispy_forms.layout import Layout, Div, Fieldset
 
 
-tns_search_url = 'https://www.wis-tns.org/api/get/search'
-tns_object_url = 'https://www.wis-tns.org/api/get/object'
+TNS_BASE_URL = 'https://www.wis-tns.org/'
+TNS_OBJECT_URL = f'{TNS_BASE_URL}api/get/object'
+TNS_SEARCH_URL = f'{TNS_BASE_URL}api/get/search'
 
 
 class TNSForm(GenericQueryForm):
@@ -102,7 +103,7 @@ class TNSBroker(GenericBroker):
                 'public_timestamp': public_timestamp,
             })
          }
-        response = requests.post(tns_search_url, data)
+        response = requests.post(TNS_SEARCH_URL, data)
         response.raise_for_status()
         transients = response.json()
         alerts = []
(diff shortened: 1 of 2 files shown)


tom_alerts/brokers/tns.py
```diff
@@ -7,8 +7,8 @@
 from crispy_forms.layout import Layout, Div, Fieldset
 
 
-tns_search_url = 'https://wis-tns.weizmann.ac.il/api/get/search'
-tns_object_url = 'https://wis-tns.weizmann.ac.il/api/get/object'
+tns_search_url = 'https://www.wis-tns.org/api/get/search'
```
dmcollom, line 10: So, ideally, to avoid code duplication, this would look a little more like this:

```
TNS_BASE_URL = 'https://www.wis-tns.org/'
TNS_OBJECT_URL = f'{TNS_BASE_URL}api/get/object'
TNS_SEARCH_URL = f'{TNS_BASE_URL}api/get/search'
```

This way, it only need to be changed in one place in the future. Constants in all caps is also Python convention.

tom_alerts/brokers/tns.py
```diff
@@ -138,7 +138,7 @@ def fetch_alerts(cls, parameters):
     def to_generic_alert(cls, alert):
         return GenericAlert(
             timestamp=alert['discoverydate'],
-            url='https://wis-tns.weizmann.ac.il/object/' + alert['objname'],
+            url='https://www.wis-tns.org/object/' + alert['objname'],
```
dmcollom, line 141: Just out of curiosity, because I don't have a TNS API key to test this, is this the correct URL? The object url listed above is at `/api/get/object/`--just want to be sure.

dmcollom (changes requested):
Thanks for noticing this and correcting it! Just looking for a quick stylistic change.

Also, are you having any issues using the module? I tried to make the base url change a while back, and (with a borrowed API key) wasn't able to successfully make any queries.

tom_alerts/brokers/tns.py
griffin-h: Yes I think the idea here was to have the alert URL go to a human-readable page, rather than the API. For example, https://www.wis-tns.org/object/2020mot.

tom_alerts/brokers/tns.py
griffin-h: Sounds good. I was just updating what was already there (which apparently I also wrote, so no wonder it's bad style), but I've fixed this now.

griffin-h:
> Also, are you having any issues using the module? I tried to make the base url change a while back, and (with a borrowed API key) wasn't able to successfully make any queries.

It seems to work for me, but maybe someone in the supernova group there can confirm?

dmcollom (approved):

PR #414, opened 2021-02-02
use energy flux in SpectrumSerializer

griffin-h:
Currently the `SpectrumSerializer` stores spectra in the form of `photon_flux` (units of photons/s/cm^2/Å or similar). As far as I know, almost all optical and infrared astronomers use energy flux (units of erg/s/cm^2/Å or similar), including the `SpectroscopyProcessor.DEFAULT_FLUX_CONSTANT`. When it deserializes, it stores the photon flux as the `flux` of the new `Spectrum1D` object. That means that the `Spectrum1D` object does not survive the round-trip of `serialize` and `deserialize`.

Was there some reason for using photon flux in the first place? If not, I would suggest simply storing the `flux` attribute of the `Spectrum1D` object, which will have whatever units are input by the user. This should avoid confusion.


griffin-h:
I hope @cmccully will check that I'm not totally crazy here!

dmcollom:
I'm gonna let @cmccully handle this one but I will mention that the tests are currently failing--those will need to pass for this PR to be approved.

cmccully:
It looks like we need to rename "photon_flux" in the tests as well to get them to pass.

griffin-h:
> It looks like we need to rename "photon_flux" in the tests as well to get them to pass.

Done.

griffin-h:
I'll also mention that there should probably be some warning to users about this. If anyone has uploaded a spectrum to their TOM before, this is a potentially breaking change.

dmcollom:
So I'm going to merge this, and then open a PR for a migration script that will correct the data for TOMs that have been using this code. Hopefully a release early next week.

PR #416, opened 2021-02-08
Feature/convert to jsonfield

dmcollom:


phycodurus (approved):
we might think about adding type hints as we go along

phycodurus (approved):

PR #430, opened 2021-03-05
Feature/user registration

dmcollom:
This pull request breaks out user list into a few template tags in order to support `tom_registration` and includes documentation to support the new app. It also includes a few documentation updates that I noticed along the way.


phycodurus (approved):
I deleted `tom_common/templates/auth/partials/active_users_list.html`

PR #437, opened 2021-03-11
Feature/cancel obs api

dmcollom:
Added cancel API endpoint and added groups support to observation submission.


phycodurus (approved):
Nice release notes

PR #442, opened 2021-03-29
Added check to exclude spectra that don't need conversions

dmcollom:


phycodurus (approved):
Please, please add some context. I don't think anyone will read this migration code, but if they do, they're not going to have a clue.

PR #468, opened 2021-06-25
Add test for broker module import failure

cjoldelo:
All the proposed additions for MuSCAT support, as well as miscellaneous additions for comment.


dmcollom:
> @dmcollom When I run this test code, it returns the correct import error but still returns the test as failed. This morning you mentioned creating a fake response for the code to return; do I need to do that?
> Unittest mock traceback ex: https://docs.python.org/3/library/unittest.mock.html#quick-guide

@cjoldelo You're going to want to use [assertRaises()](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) within another context manager. Essentially, this will fail the test *unless* an exception is raised.

dmcollom:
> @dmcollom I saw that my commit from a few days ago failed a publish_coverage test with the exception:
> 'coveralls.exception.CoverallsException: Running on Github Actions but GITHUB_TOKEN is not set. Add "env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}" to your step config.'
> How can I resolve this exception?
> Thanks!

So (I believe) this is happening because your repo is a fork of ``tom_base``, and you need to set a Github secret value with the correct token. I will send you the token via Slack, and you should set the proper secret in your forked repo.

cjoldelo:
@dmcollom This Pull Request is ready for review!

test_attempt.py
```diff
@@ -0,0 +1,38 @@
+import unittest
```
dmcollom, line 1: In general, this test should live in `tom_base/tom_alerts/tests/tests.py`, and grouped in a more appropriate spot there, rather than just living on its own.

test_attempt.py
```diff
@@ -0,0 +1,38 @@
...
+from django.urls import reverse
+
+from tom_alerts.alerts import GenericBroker, GenericQueryForm, GenericUpstreamSubmissionForm, GenericAlert
+from tom_alerts.alerts import get_service_class
+from tom_alerts.exceptions import AlertSubmissionException
+from tom_alerts.models import BrokerQuery
+from tom_observations.models import ObservationRecord
+from tom_targets.models import Target
+
+# env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dmcollom, line 22: This should be removed.

test_attempt.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Test alert data. Normally this would come from a remote source.
```
dmcollom, line 24: If this test data isn't used, it should also be removed.

test_attempt.py
```diff
@@ -0,0 +1,38 @@
...
+    {'id': 1, 'name': 'Tatooine', 'timestamp': '2019-07-01', 'ra': 32, 'dec': -20, 'mag': 8, 'score': 20},
+    {'id': 2, 'name': 'Hoth', 'timestamp': '2019-07-02', 'ra': 66, 'dec': 50, 'mag': 3, 'score': 66},
+]
+
+class TestModuleImport(TestCase):
+
+#Failure to import module
+    def test_import_invalid_mod(self):
+        with self.subTest('Test that an invalid import returns an import error.'):
+            with patch('tom_observations.api_views.get_service_class') as mock_get_service_class:
```
dmcollom, line 35: I'm a little unsure that this is testing the proper item. You're intending to test `tom_alerts.brokers.broker.get_service_class`, but this is patching `tom_observations.api_views.get_service_class`. This looks like a copy/paste job that should be corrected, and ensured to be testing the correct item.

test_attempt.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Test alert data. Normally this would come from a remote source.
+test_alerts = [
+    {'id': 1, 'name': 'Tatooine', 'timestamp': '2019-07-01', 'ra': 32, 'dec': -20, 'mag': 8, 'score': 20},
+    {'id': 2, 'name': 'Hoth', 'timestamp': '2019-07-02', 'ra': 66, 'dec': 50, 'mag': 3, 'score': 66},
+]
+
+class TestModuleImport(TestCase):
+
+#Failure to import module
```
dmcollom, line 32: So with tests, the proper way to comment is to put a string under the method signature, and then the test description will print out to console during a verbose run. For example in this case:

```python
def test_import_invalid_mod(self):
    """Test that attempting to import a nonexistent broker module raises an ImportError"""
    ...
```

Also, a couple _really_ nitpicky notes on that:
- Comments should be at the same depth as what they're commenting--so this one, which will ultimately be replaced by the string below the signature, would normally be indented to match `def test_import_valid_mod`
- Comments need a space between the octothorpe and the comment itself

test_attempt.py
```diff
@@ -0,0 +1,38 @@
...
+
+# env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+# Test alert data. Normally this would come from a remote source.
+test_alerts = [
+    {'id': 1, 'name': 'Tatooine', 'timestamp': '2019-07-01', 'ra': 32, 'dec': -20, 'mag': 8, 'score': 20},
+    {'id': 2, 'name': 'Hoth', 'timestamp': '2019-07-02', 'ra': 66, 'dec': 50, 'mag': 3, 'score': 66},
+]
+
+class TestModuleImport(TestCase):
```
dmcollom, line 30: As kinda stated above, this should probably be consolidated with a more appropriate test class.

test_attempt.py
```diff
@@ -0,0 +1,38 @@
...
+
+#Failure to import module
+    def test_import_invalid_mod(self):
+        with self.subTest('Test that an invalid import returns an import error.'):
+            with patch('tom_observations.api_views.get_service_class') as mock_get_service_class:
+                mock_get_service_class.side_effect = ImportError('Import failed. Did you provide the correct path?')
+                with self.assertRaises(ImportError):
+                    mock_get_service_class()
```
dmcollom, line 38: Can you clarify what we're trying to test here, again? Because what it looks like is that you're creating a mock object, telling it to raise an Exception, and then calling it and making sure it raises the Exception. It doesn't look like this ever calls any real TOM Toolkit code, so it isn't really testing anything.

dmcollom (changes requested):

test_attempt.py
cjoldelo: You're right that this is an incorrect copy/paste, though I don't even remember where I got the example from. It should be tom_alerts.alerts.get_service_classes, which I will change.

test_attempt.py
cjoldelo: This is for two specific lines of code in the get_service_classes function in tom_alerts.alerts that Coveralls said wasn't tested. I thought I would start small for the unit testing. The lines of code are:
except (ImportError, AttributeError):
[indent] raise ImportError(f'Could not import {service}. Did you provide the correct path?')
I thought that in order to test the code, you just had to create an ImportError and see that the code returned what you told it to. Does the mocking mean that I worked around the problem rather than testing the actual code?

dmcollom: Yeah, you did more or less work around the problem and not test the code. I think the patch line you copied is the one I used as an example.

If [this](https://coveralls.io/builds/40896872/source?filename=tom_alerts%2Falerts.py#L49) is the line of code you're testing, then you'd want to trigger [this line](https://coveralls.io/builds/40896872/source?filename=tom_alerts%2Falerts.py#L47) to raise the `ImportError`, and so I believe your patch should actually be `with patch('tom_alerts.alerts.import_module') as mock_import_module`, with the appropriate changes due to the variable renaming.

Something else you could do to enhance this is to add another subtest that raises an `AttributeError` instead, to ensure that it gets triggered in both cases. Also, you may consider testing that the message is what you expect, by passing a `msg` kwarg as shown [in the docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex). That will ensure that you're actually getting into the code as expected.

cjoldelo: Alright, I'll do my best to get to this tomorrow. Thank you, and sorry for all the hassle!

tom_alerts/tests/tests.py
```diff
@@ -101,6 +101,25 @@ def test_to_generic_alert(self):
     def test_to_target(self):
         target, _, _ = TestBroker().to_generic_alert(test_alerts[0]).to_target()
         self.assertEqual(target.name, test_alerts[0]['name'])
+        
+
+@override_settings(TOM_ALERT_CLASSES=['tom_alerts.fake_broker'])
```
cjoldelo, line 106: @dmcollom Let me know how you feel about the placement/if you have any more comments!

dmcollom: That looks like a fine spot--I would just rename the class to ``TestAlertModule`` and the test itself to ``test_get_service_classes_import_error``, as those are more descriptive and closer to what's actually being tested. Otherwise, I think this can be un-drafted.

dmcollom: Ah, there are a couple of linting things to be fixed, as well.

cjoldelo: @dmcollom I think it should pass the lint now. I'm having trouble running the tests in tom_base on my own machine; something about the line endings in Windows and Unix not being the same. I'm looking into it now, but apologies in advance if the code still doesn't pass the lint.

cjoldelo: I'm also still getting the Report Coverage issue from before, though I put in the Coveralls Repo Token you sent me: "Running on GitHub Actions but GITHUB_TOKEN is not set. Add GITHUB_TOKEN to your step config." Is this something to be concerned about?

dmcollom (approved):

PR #472, opened 2021-07-14
add visibility and moon separation functions and tests

cjoldelo:
Comments/suggestions welcome! This pull request is meant to add visibility and moon separation functions, plus their associated tests, to the tom. Feel free to suggest file or function name changes as well as code suggestions.


tom_observations/moon_separation.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Determines the min and max separations of the target object and the moon over a full
+    observing night at the desired observatory. If it registers <15 degree separation at
+    minimum, it prints a warning that the target object is too close to the moon.
+    If it registers a <15 degree separation at maximum, the observation request is rejected.
+    """
+    try:
+        coords = SkyCoord(ra*u.deg, dec*u.deg, frame='icrs')
+
+        try:
```
cjoldelo, line 21: @dmcollom the goal of these 4 lines (21-24) is to check if the target is listed in the Sesame Name Resolver (SIMBAD + Ned + Vizier), since those coordinates tend to be slightly more accurate. If the target is not listed, the code simply creates a FixedTarget object from the given coordinates and name. However, the Style Checks in the lint seem to take issue with these lines, since only one 'target' definition is ever used at a time (error code line 24, "local variable 'target' is assigned to but never used"). Any suggestions on how to resolve this?

cjoldelo: Never mind! Turns out I've changed this code so much that I never end up using 'target,' so I'll just take it out. Thanks anyway!

dmcollom:
I'm wondering what the specific use case is for the two primary functions here, `all_night_moon_sep()` and `calculate_visibility`. The Toolkit already has a couple of utilities that are similar, if not identical:

`get_sidereal_visibility()` is used to get all visibilities at all eligible sites in the Toolkit: https://github.com/TOMToolkit/tom_base/blob/48c51ab577dc9ee335def7a18ac241bf242e1da8/tom_observations/utils.py#L13

`moon_distance()` renders a plot of moon distance: https://github.com/TOMToolkit/tom_base/blob/48c51ab577dc9ee335def7a18ac241bf242e1da8/tom_targets/templatetags/targets_extras.py#L125

Now, I recognize that these functions are a little less convenient to call than what you wrote, but it's not clear to me how yours are being used. I'm certainly open to doing some refactoring of the existing functions to make them more flexible or usable, but can you let me know where the shortcomings are that necessitate this PR?

rachel3834:
I think we should be careful to distinguish between the Toolkit's
implementation and the code Claire has developed which was designed to
implement the microlensing science use-case specifically.  The refactoring
here was implemented in a way that makes it easier to use the visibility
and lunar separation data for target selection and observation request
design in the management command that runs our observing program
automatically.   As opposed to the current Toolkit implementation which is
geared towards interactions through the UI.

dmcollom:
> I think we should be careful to distinguish between the Toolkit's implementation and the code Claire has developed which was designed to implement the microlensing science use-case specifically. The refactoring here was implemented in a way that makes it easier to use the visibility and lunar separation data for target selection and observation request design in the management command that runs our observing program automatically. As opposed to the current Toolkit implementation which is geared towards interactions through the UI.

Well, yes, I understand that, which is why I was hoping to clarify the use case and discuss a potential refactor. I would not consider Claire's work to be "refactoring" as you say, but I think it can/should be made into a refactor. Both the existing and new functionality should both use shared functions that calculate visibility and moon distance, rather than introducing potentially duplicate code that now has to be maintained in the future. I think what Claire has put forth here is fine code for what MOP needs, but I do not believe it should be considered for the base Toolkit unless there's a significant refactor to unify the functionality that it duplicates.

rachel3834:
Sure, that's fine - let's talk over the refactoring at the TOM meeting
tomorrow.

dmcollom:
Closed this as it has been determined that we need to think about this quite a bit more.

PR #476, opened 2021-08-04
Adds support for Lasair's next gen broker Iris

Fingel:
The LasairIris module works by querying a "Stream Log" stored on Lasair
Iris. A steam log is a log from a stored query on lasair that is updated
in near real time and written to a static file, so no sql needs to be
run when the user requests it.

I need to support this in a TOM I'm building for another project so I
figured I'd send it upstream.

### DIFF
--- a/tom_alerts/alerts.py
+++ b/tom_alerts/alerts.py
@@ -24,6 +24,7 @@ DEFAULT_ALERT_CLASSES = [
     'tom_alerts.brokers.gaia.GaiaBroker',
     'tom_alerts.brokers.scimma.SCIMMABroker',
     'tom_alerts.brokers.fink.FinkBroker',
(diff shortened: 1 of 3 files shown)


dmcollom (changes requested):
Thanks for this! This looks great, but would you mind adding tests? It's been a slow process over the last year, but we're trying to do broker- and facility-specific tests, as well as canary tests when possible. [Alerce example](https://github.com/TOMToolkit/tom_base/blob/main/tom_alerts/tests/brokers/test_alerce.py)

Also, there was a broken test in main, but that's fixed now.

Fingel:
It was probably premature to submit this, I think the module is going to change significantly. I'll open another PR in the future if I get it to the point where it would be useful for the public.

PR #477, opened 2021-08-07
Make TOM name configurable

Fingel:
It would be nice if users would not have to override
tom_common/base.html just to update the name of the TOM that is
displayed in the header and title of each page. I think those templates
are just fine.

This introduces a backwards compatible tag that pulls a TOM_NAME setting
from settings.py but defaults to "TOM Toolkit" in case that setting
doesn't exist. I added the setting to the setup template, but it might
be a good candidate to add to the setup script as an interactive
question as well.

### DIFF
--- a/docs/common/customsettings.rst
+++ b/docs/common/customsettings.rst
@@ -294,3 +294,11 @@ Example:
       'REGISTRATION_REDIRECT_PATTERN': 'home',
       'SEND_APPROVAL_EMAILS': True
    }
+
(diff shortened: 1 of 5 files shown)


dmcollom (changes requested):
Excellent idea--can you just add the setting to `docs/common/customsettings.rst`?

Fingel:
Added the docs.

dmcollom (approved):

PR #480, opened 2021-08-09
Adds a move function for the target list

Fingel:
This PR adds functions for moving targets between TargetLists. If a
target is moved to a TargetList, it's previous lists are cleared and
the target is added to the specified TargetList. If the target already
exists in the specified TargetList a warning is displayed but the
target's previous associations are not changed.

In addition a button is added to the top of the target table alongside
the add/remove buttons.

This is a faster way of doing remove/add for a large quantity of
targets.

### DIFF
--- a/tom_targets/groups.py
+++ b/tom_targets/groups.py
@@ -28,7 +28,7 @@ def add_all_to_grouping(filter_data, grouping_object, request):
         return
     for target_object in target_queryset:
         try:
-            if not request.user.has_perm('tom_targets.view_target', target_object):
+            if not request.user.has_perm('tom_targets.change_target', target_object):
                 failure_targets.append((target_object.name, 'Permission denied.',))
             elif target_object in grouping_object.targets.all():
                 warning_targets.append(target_object.name)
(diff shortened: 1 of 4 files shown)


tom_targets/groups.py
```diff
@@ -167,3 +167,92 @@ def remove_selected_from_grouping(targets_ids, grouping_object, request):
...
+    failure_targets = []
+    try:
+        target_queryset = TargetFilter(request=request, data=filter_data, queryset=Target.objects.all()).qs
+    except Exception:
+        messages.error(request, "Error with filter parameters. No target(s) were moved to group '{}'."
+                                .format(grouping_object.name))
+        return
+    for target_object in target_queryset:
+        try:
+            if not request.user.has_perm('tom_targets.view_target', target_object):
```
dmcollom, line 199: I think this might make more sense as the `change_target` permission. What do you think?

tom_targets/groups.py
```diff
@@ -167,3 +167,92 @@ def remove_selected_from_grouping(targets_ids, grouping_object, request):
...
+    :param request: request object passed to the calling view
+    :type request: HTTPRequest
+    """
+    success_targets = []
+    warning_targets = []
+    failure_targets = []
+    for target_id in targets_ids:
+        try:
+            target_object = Target.objects.get(pk=target_id)
+            if not request.user.has_perm('tom_targets.view_target', target_object):
```
dmcollom, line 241: Same question for you as above.

dmcollom (approved):
Looks good! Just had the one question about your thoughts on the appropriate permission.

Fingel:
> Looks good! Just had the one question about your thoughts on the appropriate permission.

Yes, it does make more sense to use change_target I think. I took the liberty and also changed it for the other methods I copied and pasted 😂

PR #483, opened 2021-08-12
Fix query for the "Recent Photometry" table

Fingel:
The query doesn't actually filter by target, so it does not reflect
recent photometry for the target displayed on the Target Detail page.


dmcollom (approved):
Gah, I've been meaning to fix this for ages, but I keep forgetting. I even had a card for it: https://github.com/TOMToolkit/tom_base/projects/1#card-39022219

Thanks!

PR #485, opened 2021-08-12
Fixes two issues with MARS photometry ingest

Fingel:
This PR adresses 2 issues with the MARS ingest:

1. The filter ids from mars are not 0 indexed, so photometry was being
   stored with incorrect filters.
2. The magnitude error was not being stored even though the photometry
   plot supports it.


Fingel:
Superceded by #487

Issue #486, opened 2021-08-12
Tests do not mock out API calls for facility status


dmcollom:
**Describe the bug**
A test run failed due to a failed call to the LCO facility status API. Unit tests should not make external calls. See https://github.com/TOMToolkit/tom_base/pull/483/checks?check_run_id=3314179862

**To Reproduce**
1. Disable network.
2. Run `./manage.py test --exclude-tag=canary`
3. Tests should fail.

dmcollom:
Fixed as a side effect of #489 .

PR #487, opened 2021-08-13
Add mag error and nondetections to mars and plot

Fingel:
This PR implements a few things:

1. Adds support for importing magnitude error from MARS lightcurves.
2. Adds support for importing non detections from MARS lightcurves.
3. Implements plotting both of the above on the photometry for target
   plot.
4. Fixes photometry being saved under the wrong filter.

I hope you will consider mergint this, as by default the photometry plot
is missing quite a bit of useful data.

image

### DIFF
--- a/tom_alerts/brokers/mars.py
+++ b/tom_alerts/brokers/mars.py
@@ -11,7 +11,7 @@ from tom_targets.models import Target
 from tom_dataproducts.models import ReducedDatum
 
 MARS_URL = 'https://mars.lco.global'
-filters = {0: 'g', 1: 'r', 2: 'i'}
+filters = {1: 'g', 2: 'r', 3: 'i'}
 
 
 class MARSQueryForm(GenericQueryForm):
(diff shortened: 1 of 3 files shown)


dmcollom:
I'm happy to merge it, as I agree that it's very necessary, but the tests will need to be passing first.

Fingel:
OK cool, I fixed the tests. Admittedly I didn't run them, the suite takes so long to run. Is it making network requests or something?

dmcollom:
Well, I discovered yesterday that's there's at least one test for a view that makes a network call (#486 ), but there are canary tests that run when you run a simple `./manage.py test`. The way to exclude those is with `./manage.py test --exclude-tag=canary`, which is now appropriately documented [here](https://github.com/TOMToolkit/tom_base/blob/main/README-dev.md#run-tests).

dmcollom (approved):

Fingel:
Oh nice, I'll run without the canary tests next time.

PR #488, opened 2021-08-13
Do not artificially limit alert results

Fingel:
This was probably put in here for testing and not removed, but the
results from alert brokers should not be limited, but if they must be
limited, it should be documented somewhere on the template. I think it
should be fine to just return them all since the TOM has already
downloaded them all at this point.


dmcollom (approved):
I've always wondered why it was limited to 20, but wanted to wait to change this until we add pagination...probably better to fix it now.

PR #494, opened 2021-09-01
Updates for new TNS API requirements

Fingel:
The TNS api now requires that you include a User-Agent header with
values for both your "bot name" and "bot id", and will return a 403 if
they are not provided. Why they can't just associate this info with your
API key?? Who knows.


tom_alerts/brokers/tns.py
```diff
@@ -82,6 +82,15 @@ class TNSBroker(GenericBroker):
     name = 'TNS'
     form = TNSForm
 
+    @classmethod
+    def tns_headers(cls):
```
dmcollom, line 86: Can you just comment a link to the documentation where this is listed? I'm assuming it's the ``API - Search/Get objects`` section [here](https://www.wis-tns.org/content/tns-getting-started). And just to double-check, the documented code has a single pair of braces as opposed to the double pair--does it work either way?

dmcollom (approved):
Sorry for the delay--was on vacation last week. Just the one request for the documentation, and Lindy and I unfortunately muddied the PR a little bit by changing the target branch. Thanks for the change!

phycodurus (approved):

tom_alerts/brokers/tns.py
Fingel: I actually found it in some obscure news article they posted:
https://www.wis-tns.org/content/tns-newsfeed#comment-wrapper-23710
It's not in the actual API docs from what I can tell: https://www.wis-tns.org/sites/default/files/api/TNS_APIs_manual.pdf

Fingel: For the record - the double braces is to escape the braces, the final string only has one.

PR #499, opened 2021-09-17
Add a few configuration options to the photo plot

Fingel:
This PR adds a few config options to the photometry plot, so that some
basic aesthetic parameters can be changed. Width, height, background
color, etc. Defaults should show no change.

I'm not sure the best place to document this. Any thoughts?


dmcollom:
> This PR adds a few config options to the photometry plot, so that some
> basic aesthetic parameters can be changed. Width, height, background
> color, etc. Defaults should show no change.
> 
> I'm not sure the best place to document this. Any thoughts?

I would just do a more robust docstring the way it's done in some of the rest of the project, like [here](https://github.com/TOMToolkit/tom_base/blob/bd649f9dd0432d6399ce766c4750f888f8d1f4b7/tom_targets/utils.py#L15). It looks like the templatetags are part of the generated API docs, so there shouldn't be anything else to be done: https://tom-toolkit.readthedocs.io/en/stable/api/tom_dataproducts/templatetags.html

dmcollom (approved):

PR #503, opened 2021-09-22
Fixes the CSS/HTML so navbar displays properly

Fingel:
The text for dropdowns was hardcoded to black, which is not good if you
want to use an astronomer approved all dark mode theme.

Applying the correct classes and html removes the need for hardcoding
black, and I think I also fixed the intended behavior of the"acive"
class, which looks a lot better now.


dmcollom (approved):

dmcollom:
This is released in 2.9.1.

PR #517, opened 2021-11-01
Updated django-filter version for new CalVer

dmcollom:
`django-filter` has switched from semantic versioning to calendar versioning. While there's a Dependabot PR, it includes a version band of 2.4 < version < 22, which could cause issues if for some reason a version is mistakenly released in that wide band.

This PR should supercede #504 and #504 should be closed.


phycodurus (approved):
Thank you, David!

PR #519, opened 2021-11-07
Fixed simbad canary test, added exception handling to catalog search

dmcollom:
Canary tests have been failing for 16 nights. While almost all of them are due to a bizarre unreproducible specutils error, the very first one was a SIMBAD query failure. The SIMBAD script service no longer returns data for Messier objects, so the change to HD289002 should resolve the issue.

The fact that the SIMBAD harvester module will no longer return Messier objects should be noted, but the new exception handling probably means that it's not a bug.


dmcollom:
As an aside to this, I heard back from the Simbad maintainer, and Messier objects are once again queryable. Because of this, the failure test I added started failing, as it was not properly mocked. This is now fixed.

phycodurus (approved):

PR #542, opened 2022-05-05
alerce: Implement process_reduced_data.

Fingel:
Implements ALeRCEBroker.process_reduced_data to enable creating
ReducedDatums from the ALeRCE broker API.


Fingel:
Error from tests:
` ValueError: Column year failed to convert: invalid literal for int() with base 10: '\x00\x00'`
I saw this in production as well, but didn't locally. I think it has something to do with an updated library. I had to disable the visibility plot to get the target detail page to load.

phycodurus:
Hi Austin, this PR is merged into `dev`.  (I created a the `feature/alerce-datums` branch to remove an unused import that the linter was choking on. Maybe there was an easier way, but you're changes are merged to `dev` now.
-Lindy

PR #551, opened 2022-07-01
protect against none extra

brendanrmills:
If a user attempts to save a None type as an extra django will throw a database error
.`django.db.utils.IntegrityError: NOT NULL constraint failed: tom_targets_targetextra.value`
As brokers provide data on targets sometimes an element of the alert is a None that has not been cast to a string or otherwise process. If a user were to try to save the alert as a bunch of TargetExtras django would not allow the None to be put into the database. This fix checks the parameter to be saved in TargetExtra, if it is `None` type then it rewrites the value as the string `'None'`

### DIFF
--- a/tom_targets/models.py
+++ b/tom_targets/models.py
@@ -428,7 +428,7 @@ class TargetExtra(models.Model):
         Saves TargetExtra model data to the database. In the process, converts the string value of the ``TargetExtra``
         to the appropriate type, and stores it in the corresponding field as well.
         """
-        if self.value == None:
+        if self.value is None:
             self.value = 'None'
         try:
             self.float_value = float(self.value)
--- a/tom_targets/models.py
+++ b/tom_targets/models.py
@@ -448,7 +448,6 @@ class TargetExtra(models.Model):
                 self.time_value = None
         else:
             self.time_value = None
-            
         super().save(*args, **kwargs)
 
     def typed_value(self, type_val):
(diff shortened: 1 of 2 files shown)


jchate6 (changes requested):

tom_targets/models.py
```diff
@@ -428,6 +428,8 @@ def save(self, *args, **kwargs):
         Saves TargetExtra model data to the database. In the process, converts the string value of the ``TargetExtra``
         to the appropriate type, and stores it in the corresponding field as well.
         """
+        if self.value == None:
```
jchate6, line 431: Should be `if self.value is None:`

tom_targets/tests/tests.py
```diff
@@ -318,6 +318,21 @@ def test_target_save_programmatic_extras(self):
...
+        target.save(extras={'foo': None})
+        self.assertTrue(TargetExtra.objects.filter(target=target, key='foo').exists())
+
+    def test_datetime_warning(self):
+        '''Tests for an int that might come in and get mistakenly parsed for a datetime
+        If the value can be successfully cast as a float it is not useful to us as a datetime
+        Casting a datetime from an int will indicate that year on an arbitrary day. '''
+        target = SiderealTargetFactory.create()
+        target.save(extras={'foo': '1984'})
+        te = target.targetextra_set.get(key = 'foo')
```
jchate6, line 332: Remove spaces around the `=` for explicit keyword arguments.

jchate6 (changes requested):

jchate6:
Lint is unhappy...

jchate6 (approved):

jchate6:
I'm not sure why this test is failing. I think this might have to wait for @phycodurus

PR #554, opened 2022-08-01
remove SOARFacility.get_facility_status inheritance

griffin-h:
Right now the "facility status" tab on the target page lists every LCO site twice. This is because the `SOARFacility` inherits its `get_facility_status` method from `LCOFacility`, which gets the status of every LCO site from the scheduler. This fix just removes that inheritance so that each LCO site is only listed once (from `LCOFacility.get_facility_status`).

### DIFF
--- a/tom_observations/facilities/lco.py
+++ b/tom_observations/facilities/lco.py
@@ -1245,34 +1245,38 @@ class LCOFacility(BaseRoboticObservationFacility):
 
         # set up the return value to be populated by the for loop below
         facility_status = {
-            'code': 'LCO',
+            'code': self.name,
             'sites': []
         }
+        site_list = [site["sitecode"] for site in self.SITES.values()]
 
         for telescope_key, telescope_value in telescope_states.items():
             [site_code, _, _] = telescope_key.split('.')
+            
+            # limit returned sites to those provided by the facility
+            if site_code in site_list:
 
-            # extract this telescope and it's status from the response
-            telescope = {
-                'code': telescope_key,
-                'status': telescope_value[0]['event_type']
-            }
-
-            # get the site dictionary from the facilities list of sites
-            # filter by site_code and provide a default (None) for new sites
-            site = next((site_ix for site_ix in facility_status['sites']
-                         if site_ix['code'] == site_code), None)
-            # create the site if it's new and not yet in the facility_status['sites] list
-            if site is None:
-                new_site = {
-                    'code': site_code,
-                    'telescopes': []
+                # extract this telescope and it's status from the response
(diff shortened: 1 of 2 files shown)


jchate6 (changes requested):
I'm not a fan of this change. I think it makes more sense to update the LCO facility to only include sites listed in the facility.

I'll work on making some commits that accomplish this.

griffin-h:
Just to clarify, SOAR is *not* listed in the LCO facility status API. (Why this is is a separate question.) So I don't think it makes sense for the `SOARFacility` to query that API at all. On the other hand, all the sites included in that API *do* belong to the `LCOFacility`, so no checking is needed there unless that changes.

jchate6:
> Just to clarify, SOAR is _not_ listed in the LCO facility status API. (Why this is is a separate question.) So I don't think it makes sense for the `SOARFacility` to query that API at all. On the other hand, all the sites included in that API _do_ belong to the `LCOFacility`, so no checking is needed there unless that changes.

So, this doesn't seem to be true. Querying `https://observe.lco.global/api/telescope_states/` returns 
```    "sor.doma.4m0a": [
        {
            "telescope": "sor.doma.4m0a",
            "event_type": "AVAILABLE",
            "event_reason": "Available for scheduling",
            "start": "2022-09-21T18:30:42Z",
            "end": "2022-09-21T18:30:42Z"
        }
    ],
```
However, the `SOR` site is NOT listed in `LCOFacility.sites` and shouldn't be included in the base call unless `sites` is otherwise defined.

I believe the changes I made address this and accomplish the ultimate goal of this PR to keep things from repeating in the observatory status list.

jchate6:
@griffin-h If you want to have a chat about this PR, just let me know.

griffin-h:
> So, this doesn't seem to be true. Querying `https://observe.lco.global/api/telescope_states/` returns
> 
> ```
>         {
>             "telescope": "sor.doma.4m0a",
>             "event_type": "AVAILABLE",
>             "event_reason": "Available for scheduling",
>             "start": "2022-09-21T18:30:42Z",
>             "end": "2022-09-21T18:30:42Z"
>         }
>     ],
> ```
> 
> However, the `SOR` site is NOT listed in `LCOFacility.sites` and shouldn't be included in the base call unless `sites` is otherwise defined.
> 
> I believe the changes I made address this and accomplish the ultimate goal of this PR to keep things from repeating in the observatory status list.

Hah, this was literally not true yesterday, and I just asked @cmccully about it. Maybe it's only listed on AEON nights (like tonight)?  Anyway I agree that your changes are better!

PR #555, opened 2022-08-05
display only relevant visibility on observation form

griffin-h:
Right now, the visibility chart on the observation form page shows the airmass for all available facilities. This quickly gets crowded, with all the LCO sites plus Gemini by default. Some of the sites (e.g., Chile) also overlap each other.

These changes restrict that chart to only plot the visibility for the facility for which you are requesting observations. The `observation_plan` function already had an argument `facility` that was not used (I assume this was the intention), which I just had to pipe through to `get_sidereal_visibility`. The default behavior of these functions has not been changed.


jchate6 (approved):
Love this. Thanks!

PR #556, opened 2022-08-05
fix sexagesimal coordinate formatting

griffin-h:
This is a minor fix to the sexagesimal coordinate formatting on the target detail page. The current specification of `05.3f` does not add leading zeros to the last field as required (it would need to be `06.3f` to count the decimal point in the string length). However, while I was here, I switched to the built-in Astropy formatting of coordinates, which is easier to understand.

I also corrected the spelling of sexagesimal in the docstring, but not in the function name, in case that would break things for users. If you wanted to, you could make an alias of this function with the correct spelling.


jchate6 (approved):
Looks great!
I'm not too concerned about the variable name for the moment.

PR #557, opened 2022-08-05
fix width of moon distance plot

griffin-h:
This is a minor fix to the observation form template so that the moon distance plot in the left column does not overlap with the right column. Screenshots (Firefox) below. I also confirmed that it now dynamically adjusts the width to the window size.

Before:
before

After:
after

### DIFF
--- a/tom_observations/templates/tom_observations/observation_form.html
+++ b/tom_observations/templates/tom_observations/observation_form.html
@@ -17,13 +17,9 @@
 {% endif %}
 <div class="row">
     <div class="col-md-6">
-        <div class="row">
-            {% target_data target %}
-        </div>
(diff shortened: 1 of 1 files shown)


jchate6 (commented):
The requested changes have the opposite of the desired effects on my machine (Fedora) on both Chrome and Firefox. 
I'll try to look into a more robust solution.

jchate6 (approved):
These changes work for me locally.
@griffin-h Does this work for you?

griffin-h:
Yes, it works and looks much more beautiful. Thanks!

Screenshots (Firefox 104.0.2 on Ubuntu 22.04.1):
image
image

Issue #561, opened 2022-09-09
Target creation tests fail for django-4.1.1


jchate6:
Target creation tests fail for Django-4.1.1 with a `ValueError`

**To Reproduce**
Run `./manage.py test tom_targets.tests.tests.TestTargetCreate.test_create_target`

**Expected behavior**
Tests pass with Django 4.0.6

**Additional context**
```======================================================================
ERROR: test_create_target (tom_targets.tests.tests.TestTargetCreate)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jchatelain/git/tom_base/tom_targets/tests/tests.py", line 202, in test_create_target
    response = self.client.post(reverse('targets:create'), data=target_data, follow=True)
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/test/client.py", line 852, in post
    response = super().post(
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/test/client.py", line 441, in post
    return self.generic(
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/test/client.py", line 541, in generic
    return self.request(**r)
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/test/client.py", line 810, in request
    self.check_exception(response)
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/test/client.py", line 663, in check_exception
    raise exc_value
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/core/handlers/base.py", line 199, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/core/handlers/base.py", line 362, in process_exception_by_middleware
    response = middleware_method(request, exception)
  File "/home/jchatelain/git/tom_base/tom_common/middleware.py", line 29, in process_exception
    raise exception
  File "/home/jchatelain/git/tom_base/env_tombase/lib64/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
[...]

PR #563, opened 2022-09-17
Fix/test fails

jchate6:
Closes issue #561.

This issue was cause by the target creation validation checking if the target name matched the linked aliases before either target or aliases were saved in the DB.

Specifically I added a check to see if a target had a primary Key before attempting to check for aliases, so this validation would only run during updates and not during creation.

We currently do not check for uniqueness of name/aliases during creation.

### DIFF
--- a/tom_targets/models.py
+++ b/tom_targets/models.py
@@ -265,6 +265,8 @@ class Target(models.Model):
         Ensures that Target.name and all aliases of the target are unique. Called automatically on save.
         """
         super().validate_unique(*args, **kwargs)
+        print(self.aliases.all())
+        print("========================================================")
         for alias in self.aliases.all():
             if alias.name == self.name:
                 raise ValidationError('Target name and target aliases must be unique')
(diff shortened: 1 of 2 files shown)


tom_targets/models.py
```diff
@@ -265,9 +265,11 @@ def validate_unique(self, *args, **kwargs):
         Ensures that Target.name and all aliases of the target are unique. Called automatically on save.
         """
         super().validate_unique(*args, **kwargs)
-        for alias in self.aliases.all():
-            if alias.name == self.name:
-                raise ValidationError('Target name and target aliases must be unique')
+        # Alias Check only necessary when updating target existing target. Reverse relationships require Primary Key.
+        if self.pk:
```
jchate6, line 269: This is the big change in this PR. 
I feel like this make more sense than a try/except ValueError because we don't want to hit DBs more than we need to.

jchate6:
We don't currently run any validation tests to make sure that there are no aliases that match the target name at the time of creation. If I understand [the docs](https://docs.djangoproject.com/en/4.1/releases/4.1/#reverse-foreign-key-changes-for-unsaved-model-instances) correctly, before this change, we would not have successfully validated this at creation anyway because the foreign key references to alias would have always returned null and never matched the target name.

The motivation for this PR was getting this code working with Django 4.1. If this additional validation is a priority, we should probably make a 2nd issue.

phycodurus (approved):

PR #566, opened 2022-09-22
Fix/lasair broker query

### LINKED ISSUE #547: Error in running Lasair broker query
**Describe the bug**
When you run a alert query using the Lasair Broker, a JSON decode error is thrown. 

**To Reproduce**
1. install TOM and run server
2. Navigate to Alerts page
3. Create new Lasair Query (cone search or SQL query both produce the same error)
4. Run query on the query list page

**Expected behavior**
When you run the Lasair alerts query, it should generate a list of alerts that should be displayed so we can select targets. 

**Screenshots**
Creating the Lasair Broker
Screenshot from 2022-06-17 17-27-10

Error Running the broker
Screenshot from 2022-06-17 17-27-28

CentOS Linux release 7.9.2009 (Core)

jchate6:
resolves #547.

### DIFF
--- a/tom_alerts/views.py
+++ b/tom_alerts/views.py
@@ -200,7 +200,7 @@ class RunQueryView(TemplateView):
         broker_class = get_service_class(query.broker)()
         # TODO: Should the deepcopy be in the brokers?
         alert_query_results = broker_class.fetch_alerts(deepcopy(query.parameters))
-        # Check if feedback is available for fetch_alerts, and allow for backwards compatability if not.
+        # Check if feedback is available for fetch_alerts, and allow for backwards compatibility if not.
         if type(alert_query_results) is tuple:
             alerts, broker_feedback = alert_query_results
         else:
--- a/tom_alerts/views.py
+++ b/tom_alerts/views.py
@@ -198,7 +198,8 @@ class RunQueryView(TemplateView):
         context = super().get_context_data()
         query = get_object_or_404(BrokerQuery, pk=self.kwargs['pk'])
         broker_class = get_service_class(query.broker)()
-        alerts = broker_class.fetch_alerts(deepcopy(query.parameters))  # TODO: Should the deepcopy be in the brokers?
+        # TODO: Should the deepcopy be in the brokers?
+        alerts, messages = broker_class.fetch_alerts(deepcopy(query.parameters))
         context['score_description'] = broker_class.score_description
         context['alerts'] = []
         query.last_run = timezone.now()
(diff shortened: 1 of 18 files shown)


jchate6:
Currently this simply disables the Lasair option without removing it. 
The entire Lasair API has been changed and now requires authentication tokens making this a more complex story than first expected.

jchate6:
Still need to do the following:

- [ ] Handle broken SQL Querry
- [ ] Remove unused object query code
- [ ] Update form to handle RA/DEC/Radius w/ cone
- [ ] Handle previously broken Queries
- [ ] Add Env lookup for token in Settings
- [ ] Add tests

jchate6:
Previously stored queries are entirely unsupported (no attempts to translate to current forms) but should not crash TOM.

tom_base/settings.py
```diff
@@ -23,6 +23,7 @@
 
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = os.getenv('SECRET_KEY', 'testkey')
+LASAIR_TOKEN = os.getenv('LASAIR_TOKEN', None)
```
jchate6, line 26: Is this the best place for this?

tom_alerts/views.py
```diff
@@ -198,9 +198,11 @@ def get_context_data(self, *args, **kwargs):
         context = super().get_context_data()
         query = get_object_or_404(BrokerQuery, pk=self.kwargs['pk'])
         broker_class = get_service_class(query.broker)()
-        alerts = broker_class.fetch_alerts(deepcopy(query.parameters))  # TODO: Should the deepcopy be in the brokers?
+        # TODO: Should the deepcopy be in the brokers?
+        alerts, messages = broker_class.fetch_alerts(deepcopy(query.parameters))
```
jchate6, line 202: This change requires a change to each broker alert return. 
I think this is OK, because this would add the infrastructure for these brokers to return errors, but I could easily bypass this with a try/except or something.

jchate6:
@phycodurus  I'd appreciate any help I can get on the best way to document these changes for TOM administrators.

tom_alerts/views.py
phycodurus: please rename `messages` to `broker_feedback` (sorry).

tom_base/settings.py
phycodurus: please add section for tom_alerts configurations like this:
```
#
# tom_alerts configuration
#
```

phycodurus (approved):

PR #568, opened 2022-10-12
Stretch dependencies and requirements

jchate6:
We want our dependencies to be as permissive as possible and eliminate `==` requirements.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     setup_requires=['setuptools_scm', 'wheel'],
     install_requires=[
         'astroquery>=0.4.2',
-        'astroplan==0.8',
+        'astroplan~=0.8',
         'astropy>=5.0',
         'beautifulsoup4~=4.9',
         'django>=3.1,<5.0',  # TOM Toolkit requires db math functions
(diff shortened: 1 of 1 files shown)


phycodurus (approved):

setup.py
```diff
@@ -29,7 +29,7 @@
     setup_requires=['setuptools_scm', 'wheel'],
     install_requires=[
         'astroquery>=0.4.2',
-        'astroplan==0.8',
+        'astroplan~=0.8',
         'astropy>=5.0',
```
cmccully, line 33: Check and see if astropy can go down to version 4 and things still work.

jchate6:
I'm running into a `psycopg2` error.
```
File "/home/jchatelain/git/tom_base/tomenv38/lib64/python3.8/site-packages/django/db/backends/postgresql/base.py", line 28, in <module>
    raise ImproperlyConfigured("Error loading psycopg2 module: %s" % e)
django.core.exceptions.ImproperlyConfigured: Error loading psycopg2 module: No module named 'psycopg2'
```

setup.py
jchate6: @cmccully 
`astropy 4.x` breaks with `python 3.10`, but more importantly, it breaks with `numpy > 1.22` without requiring `<=v1.22` as a dependency. So if we wanted to allow `astropy 4.x` we would need to restrict `numpy` ourselves.

We also get one failing test for `astropy 4.x` and `numpy  1.22`
```
FAIL: test_create_thumbnail (tom_dataproducts.tests.tests.TestDataProductModel)
Test that a failed thumbnail creation logs the correct message and does not break.
```

jchate6: I might be able to make sure the test works w/ `astropy 4.x` and `astropy 5.1` works with `numpy 1.22`
So a requirement of `astropy >= 4.1` (4.0 breaks) and `numpy <=1.22` is possible.

jchate6: I think the better path is to require python 3.8+
We already don't support tests on 3.7.

What's our highest priority here?

cmccully: Which part of the code breaks with 3.7? Is it our code or deps?

Why can't we support astropy 4.0 if you have numpy < 1.22 and astropy 5 with numpy > 1.22? I'm not fully understanding the thought process here.

PR #569, opened 2022-10-14
add absolute magnitude axis to photometry plot

griffin-h:
This adds an absolute magnitude axis to the photometry plot on the target detail page, if the target distance is populated. I could not find a way to add a secondary y-axis with a fixed offset from the primary y-axis in Plotly, so I had to add some code that manually scales the axis based on the data. I also had to move the legend to the top of the plot so that it doesn't overlap with the new tick labels.


tom_dataproducts/templatetags/dataproduct_extras.py
```diff
@@ -169,6 +171,8 @@ def photometry_for_target(context, target, width=700, height=600, background=Non
                 )
             )
             plot_data.append(series)
+            all_ydata.append(np.array(filter_values['magnitude']) + np.array(filter_values['error']))
+            all_ydata.append(np.array(filter_values['magnitude']) - np.array(filter_values['error']))
```
jchate6, line 175: These will break for any malformed data (Data with Nonetype in the error or Mag) that exist in the DB. 
You're also translating types here, so you might want to set the array type to `float64` to make sure it gets handled correctly.

tom_dataproducts/templatetags/dataproduct_extras.py
```diff
@@ -180,19 +184,65 @@ def photometry_for_target(context, target, width=700, height=600, background=Non
                 name=filter_name + ' non-detection',
             )
             plot_data.append(series)
+            all_ydata.append(filter_values['limit'])
```
jchate6, line 187: Limit could also theoretically contain `None` values.

tom_dataproducts/templatetags/dataproduct_extras.py
```diff
@@ -180,19 +184,65 @@ def photometry_for_target(context, target, width=700, height=600, background=Non
...
+        'mirror': True,
+    }
+    if target.distance is not None:
+        dm = 5. * (np.log10(target.distance) + 5.)  # assumes target.distance is in Mpc
+        yaxis2 = {
+            'title': 'Absolute Magnitude',
+            'range': (ymax_view - dm, ymin_view - dm),
+            'showgrid': False,
+            'overlaying': 'y',
+            'side': 'right',
```
jchate6, line 216: you need to include a `'zerolinecolor': 'rgba(0,0,0,0)',` here to remove the "zero" line added if absmag ever goes positive (you know, for people who look at things that are ACTUALLY 10 pc away :D).

jchate6 (changes requested):

tom_dataproducts/templatetags/dataproduct_extras.py
griffin-h: Hah I literally never would have considered absolute magnitudes (or apparent magnitudes) close to zero. I just turned the lines off with `'zeroline': False`.

tom_dataproducts/templatetags/dataproduct_extras.py
griffin-h: By forcing the dtype to `float`, any `None` values will get converted to `nan`, which I then ignore using `np.nanmax` and `np.nanmin`.

tom_dataproducts/templatetags/dataproduct_extras.py
```diff
@@ -285,20 +292,68 @@ def photometry_for_target(context, target, width=700, height=600, background=Non
...
+        'range': (ymax_view, ymin_view),
+        'showgrid': grid,
+        'color': label_color,
+        'showline': True,
+        'linecolor': label_color,
+        'mirror': True,
+        'zeroline': False,
+    }
+    if target.distance is not None:
+        dm = 5. * (np.log10(target.distance) + 5.)  # assumes target.distance is in Mpc
```
jchate6, line 319: So, the `Distance` stored in the DB is in parsec according to all the Help text and documentation.

See `tom_targets.models` line 198.

jchate6 (changes requested):

tom_dataproducts/templatetags/dataproduct_extras.py
jchate6: There are ways to store unit information in a django table, but that's probably beyond the scope of this PR.

griffin-h: I agree it's beyond this PR, but it would be nice to be able to choose different units, even if it's just a global setting for your TOM. Extragalactic people will definitely assume all distances are in Mpc if they don't read the documentation.

jchate6 (approved):
Looks good!

PR #590, opened 2023-01-11
incomplete first draft of expanded data sharing docs

phycodurus:


docs/managing_data/customizing_data_sharing.rst
```diff
@@ -1,10 +1,24 @@
-Customizing Data Sharing
----------------------------
+Sharing Data with Other TOMs
+############################
 
-Currently, data sharing is only possible with HERMES.
+TOM Toolkit supports data sharing in two ways: sharing data directly from one TOM Toolkit-based TOM to another and
```
jchate6, line 4: I would REALLY like to decouple TOM-TOM from TOM-HERMES-TOM sharing. They are dramatically different things that will need to be treated entirely differently by users, admin, and developers.

I think we should break this into SENDING and RECEIVING with each of those sections giving examples for both TOM and HERMES use cases.

jchate6 (commented):
We should maybe talk about this a bit?

docs/managing_data/customizing_data_sharing.rst
jchate6: Hermes is also a very specific use case while TOM-TOM is much more generally useful for all TOMs.

docs/managing_data/stream_pub_sub.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Then you will need to add an ``ALERT_STREAMS`` configuration dictionary to your ``settings.py`` that gives the credentials
+for the various streams to which you wish to subscribe. This configuration dictionary will include a list of handlers for
+each topic so that different streams can be handled differently.
+
+Some alert handlers are included as examples. Below we demonstrate how to connect to a Hermes Topic.
+
+.. code:: python
+
+    ALERT_STREAMS = [
```
jchate6, line 58: @phycodurus Do you want to add a GCN stream here?

jchate6 (approved):

PR #593, opened 2023-01-12
Apply more robust name recognition during target and data ingestion

### LINKED ISSUE #583: Add robust Target Name Recognition when ingesting target data
When ingesting new targets (see issue #70 ) or trying to ingest photometry from a TOM or a stream such as HERMES, we want to be able to make sure that the new target does not match the names or aliases of any existing targets. 

For this we should have a robust matching function that does the following:

- [x] Check if the object name matches any target name or aliases
- [x] Performs all checks with case insensitivity and regardless of spaces/special characters
- [ ] Checks object name against trusted catalog to find other known aliases. to be checked against target names in DB

Checks should be performed at both target and Data creation

jchate6:
closes #583


jchate6:
If you take a look at the most recent test messages on `hermes.test` and `tomtoolkit.test` you'll see a bunch of target names are considered to be the same and match the Target name or alias in my local TOM.

tom_dataproducts/alertstreams/hermes.py
```diff
@@ -117,10 +117,15 @@ def hermes_alert_handler(alert, metadata):
...
-            try:
-                target = Target.objects.get(name=row['target_name'])
-            except Target.DoesNotExist:
+            if row['target_name'] != target_name:
+                target_name = row['target_name']
+                query = Target.matches.check_for_fuzzy_match(target_name)
+            if query:
+                target = query[0]
+            else:
                 continue
```
phycodurus, line 129: This might be up to the TOM developer who might write their own handler, but it seems like we should at least log something upon ignoring a `photometry_table` `row for which we don't have a fuzzy match to the `target_name`. Thoughts?

phycodurus (approved):

tom_dataproducts/alertstreams/hermes.py
jchate6: In theory that's everything they are not interested in.

jchate6: I feel like if they want to "Log new targets for review" or something it's on the individual TOM.

phycodurus: ok

tom_targets/models.py
```diff
@@ -34,6 +34,28 @@
...
+    """Search for matches amongst Target names and Aliases"""
+    def check_for_fuzzy_match(self, name):
+        """
+        Check for case-insensitive names ignoring spaces, dashes, underscore, and parentheses.
+        :param name: The string against which target names and aliases will be matched.
+        :return: queryset containing matching Targets. Will return targets even when matched value is an alias.
+        """
+        simple_name = self.make_simple_name(name)
+        matching_names = []
+        for target in Target.objects.all():
```
jnation3406, line 47: I guess the fuzzy checking of aliases makes it hard to do this in a single query... small optimization but you can do a .prefetch_related with the aliases to get them from the db in the first call, rather than getting them once for each target in the loop.

jnation3406 (commented):
Looks good, just one optimization to reduce db calls.

tom_targets/models.py
jchate6: Excellent point.
Thanks for this.

Issue #588, opened 2023-01-09
Add Airmass optimization option


jchate6:
Add ability to optimize on airmass rather than earliest.

Issue #587, opened 2023-01-09
Support Lunar Phase


jchate6:
Allow user to select a maximum lunar phase for an observation

Issue #586, opened 2023-01-09
Support for multi-target requests


jchate6:
Allow for 2 targets to be part of the same request.
Allow for nodding with 2 configurations

Issue #585, opened 2023-01-09
Support for Dithering


jchate6:
- [ ] Inside `_build_configuration` Request Dithering from LCO API and return configuration list
- [ ] Add Dithering Form Fields to LCO `Facility`
- [ ] Add Dither Tab to Display Form
- [ ] Tabs for each dither pattern for static reloads

PR #594, opened 2023-01-21
Refactored LCO facility to add new capabilities

jnation3406:
Ive tested it locally to submit spectro/imaging/muscat forms with dithering/mosaicing/other new features and it is working.


jnation3406:
The tests should all be fixed now, and I've updated the SOAR forms to use the new LCO forms, and added in backwards compatibility for loading old ObservationRecords so existing cadences made with the old forms should keep working.

tom_observations/facilities/lco.py
```diff
@@ -123,16 +139,21 @@ def make_request(*args, **kwargs):
 
 
 class LCOBaseForm(forms.Form):
```
jnation3406, line 141: This has been refactored to only have utility methods getting details from /api/instruments. The methods have been expanded to get optical elements and generic modes

tom_observations/facilities/lco.py
```diff
@@ -174,84 +234,515 @@ def proposal_choices():
...
-    """
-    name = forms.CharField()
-    ipp_value = forms.FloatField(label='Intra Proposal Priority (IPP factor)',
-                                 min_value=0.5,
-                                 max_value=2,
-                                 help_text=ipp_value_help)
-    start = forms.CharField(widget=forms.TextInput(attrs={'type': 'date'}))
-    end = forms.CharField(widget=forms.TextInput(attrs={'type': 'date'}),
-                          help_text=end_help)
+class LCOTemplateBaseForm(LCOBaseForm):
```
jnation3406, line 237: This is the base form for the Templates, which I haven't touched yet. It only has a few fields that can be set.

tom_observations/facilities/lco.py
```diff
@@ -174,84 +234,515 @@ def proposal_choices():
...
-        self.helper.layout = Layout(
-            self.common_layout,
-            self.layout(),
-            self.button_layout()
+        self.fields['proposal'] = forms.ChoiceField(choices=self.proposal_choices())
+        self.fields['filter'] = forms.ChoiceField(choices=self.filter_choices())
+        self.fields['instrument_type'] = forms.ChoiceField(choices=self.instrument_choices())
+
+
+class AdvancedExpansionsLayout(Layout):
```
jnation3406, line 250: There are several Layout classes here that are used to compose the crispy forms layout with multiplicity and with different behavior for different instrument type form classes.

tom_observations/facilities/lco.py
```diff
@@ -174,84 +234,515 @@ def proposal_choices():
...
+                    f'c_{config_instance}_ic_{instance}_rotator_angle',
+                    css_class='col'
+                ),
+                css_class='form-row'
+            ),
+            *self._get_oe_groups_layout(config_instance, instance, oe_groups)
+        )
+
+
+class LCOBaseObservationForm(BaseRoboticObservationForm, LCOBaseForm):
```
jnation3406, line 696: This is the common fields that are in both the new form, and in the old style form. Basically stuff that is at the Request or RequestGroup level.

tom_observations/facilities/lco.py
```diff
@@ -343,18 +835,198 @@ def _build_target_fields(self):
...
+                }
+            ]
+        }
+        if self.cleaned_data.get('period') and self.cleaned_data.get('jitter') is not None:
+            payload = self._expand_cadence_request(payload)
+
+        return payload
+
+
+class LCOOldStyleObservationForm(LCOBaseObservationForm):
```
jnation3406, line 922: This is the base class of the old style forms, which is now the ImagingSequence and SpectrographSequence forms, which I haven't touched.

tom_observations/facilities/lco.py
```diff
@@ -366,54 +1038,364 @@ def _build_instrument_config(self):
 
         return [instrument_config]
 
-    def _build_acquisition_config(self):
-        acquisition_config = {}
 
-        return acquisition_config
+class LCOFullObservationForm(LCOBaseObservationForm):
```
jnation3406, line 1042: This is the base class and Imaging variant of the new style observation form

tom_observations/facilities/lco.py
```diff
@@ -435,284 +1419,242 @@ def observation_payload(self):
...
 
+    def form_name(self):
+        return 'image'
 
-class LCOMuscatImagingObservationForm(LCOBaseObservationForm):
+    def configuration_type_choices(self):
+        return [('EXPOSE', 'Exposure'), ('REPEAT_EXPOSE', 'Exposure Sequence')]
+
+
+class LCOMuscatImagingObservationForm(LCOFullObservationForm):
```
jnation3406, line 1454: The muscat and spectrograph forms just override a few fields to customize their forms and add in any fields specific to those instrument types. It relies heavily on functionality in the base LCOFullObservationForm

tom_observations/views.py
```diff
@@ -247,10 +247,17 @@ def get_initial(self):
     def post(self, request, *args, **kwargs):
         form = self.get_form()
         if form.is_valid():
-            return self.form_valid(form)
+            if 'validateButton' in request.POST:
```
jnation3406, line 250: This stuff is just the magic to hook in the validate button showing a message on successful validation.

jchate6:
Resolves a Bunch of issues:
Closes #584 
Closes #587 
Closes #585
Closes #586
Closes #588
Closes #595

phycodurus (approved):

tom_observations/facilities/lco.py
```diff
@@ -366,54 +1037,364 @@ def _build_instrument_config(self):
...
+            self.fields[f'c_{j+1}_max_lunar_phase'] = forms.FloatField(
+                help_text=max_lunar_phase_help, min_value=0, max_value=1.0, label='Maximum Lunar Phase', required=False)
+            self.fields[f'c_{j+1}_fractional_ephemeris_rate'] = forms.FloatField(
+                min_value=0.0, max_value=1.0, label='Fractional Ephemeris Rate',
+                help_text='Value between 0 (Sidereal Tracking) '
+                'and 1 (Target Tracking). If blank, Target Tracking.',
+                required=False
+            )
+            self.fields[f'c_{j+1}_target_override'] = forms.ChoiceField(
+                choices=self.target_group_choices(),
```
jchate6, line 1137: What's the motivation for restricting options to targets in the same group?
I'm not sure how everyone uses target groups, but I'm not sure it would make sense in many cases to use them in this fashion. 
For example, If I wanted a FLOYDS spectrum of a non-sidereal target, I'd want to observe the target (Maybe in a group `CRITICAL` or `RADAR`) and then a nearby standard (Probably in a group `STANDARD`). In order to have both listed I'd either need to make a special group for every observation or just include `STANDARDS` in my `RADAR` and `CRITICAL` groups which undermines their utility.

It would probably be better to just put a search field here similar to what we have in `tom_targets.filters`. Could throw an error or bring up a modal if it produces multiple targets upon validation.

I also see how this could be a "not now" problem.

jchate6: I assume we can't have one dropdown = `Group` and another that populates with the related `Targets` without Json?

PR #597, opened 2023-01-30
update TNS harvester with new TNS API protocol

griffin-h:
The TNS now requires additional headers for access to its APIs. These were already implemented in the TNS Broker, but not yet in the TNS Harvester.


phycodurus (approved):

jchate6 (approved):
Thanks for this!

PR #603, opened 2023-02-02
601 alerce query fails with error from tom demo dev

### LINKED ISSUE #601: Alerce query fails with error from tom-demo-dev
**Describe the bug**
Running an Alerce query (query 10, Galactic Bulge query) from tom-demo-dev's UI produces the following error:
400 Client Error: BAD REQUEST for url: https://api.alerce.online/ztf/v1/objects/?count=false&order_by=None&order_mode=None&page=1&page_size=20&ra=17.5&dec=-27.0&radius=60

**To Reproduce**
1. Go to tom-demo-dev, click Alerts
2. Run the Alerce query labeled 'Galactic Bulge' (10 in list)

**Expected behavior**
I expect the query to return a list of targets or an empty list indicating there were no targets found.

- OS: MacOS Ventura 13.1
- Browser Chrome

**Additional context**
Would be nice to fix, but not super urgent.

phycodurus:
Previously saved Broker Queries will still fail and they should be deleted and recreated.


rachel3834 (approved):
These changes look good and I think the default options available make sense.

phycodurus:
This needs to go out now. So, instead of this PR. We'll have a new one that branches from and merges to `main` (b/c this branched from `dev` and some of those changes are not ready to go out.

PR #608, opened 2023-02-07
fix alias bugs

### LINKED ISSUE #605: Alias matching is causing bugs when updating target.
**Describe the bug**
Updating Targets and aliases fails in several new and exciting ways.

From the `Update Target` form
- [x] Updating a target with an Alias breaks (alias is checked against itself)
- [x] Updating an alias name that ALREADY matches a target breaks during target validation (`Target name and target aliases must be different`)
- [x] Need to be able to check if an alias is being changed and compare new name against everything BUT the old name.

jchate6:
I think this resolves all the weird issues with existing data.

And adds tests to that effect.

Closes #605


jnation3406 (approved):

phycodurus (approved):
yay!

PR #652, opened 2023-05-10
Let Facility modules specify a custom template and add context_data

phycodurus:


jchate6 (changes requested):
This has some breaking tests.

tom_observations/views.py
```diff
@@ -198,6 +220,12 @@ def get_context_data(self, **kwargs):
 
         target = Target.objects.get(pk=self.get_target_id())
         context['target'] = target
+
+        # allow the Facility class to add data to the context
+        facility_class = self.get_facility_class()
+        facility_context = facility_class().get_facility_context_data()
```
jchate6, line 226: This line breaks the tests.

jchate6 (changes requested):

tom_observations/views.py
phycodurus: ack! was passling locally, but PR was missing the commit below

jchate6 (approved):
looks good now

PR #664, opened 2023-05-22
allow nondetections in photometry uploads

griffin-h:
This removes masked values before storing uploaded photometry data, so that some rows in the table can be detections and others can be nondetections.

I also piped through an optional `'source'` column to the `source_name` attribute of the `ReducedDatum`.

I modified the `photometry_sample.csv` file to show examples of both these changes.


jchate6 (approved):
Looks great! Thanks for this.

PR #671, opened 2023-06-30
Fix/update targets

jchate6:
Add a test for alternate strict match manager.

Fix `validate_unique` check to properly test ID value.


jnation3406 (approved):
I don't really understand all this but looks like you know what your doing

PR #674, opened 2023-07-06
Fix the date field on the tns broker

jnation3406:
Can't use a DateTimeField since we are storing the parameters in a JSONField and datetimes are not json serializable. Just a string date makes more sense anyway since we pass it through to TNS and they can interpret it how they want.


jchate6 (approved):
Looks Good!

PR #684, opened 2023-09-07
Handle 502 from telescope_states endpoint for OCS facility status info

jnation3406:


tom_observations/tests/facilities/test_ocs.py
```diff
@@ -224,6 +224,15 @@ class TestOCSFacility(TestCase):
     def setUp(self):
         self.lco = OCSFacility()
 
+    @patch('tom_observations.facilities.ocs.make_request')
+    def test_get_requestgroup_id(self, mock_make_request):
+        mock_response = Response()
+        mock_response._content = str.encode('ConnectionError - Error retrieving telescope availability')
```
mgdaily, line 230: Minor, but if we're checking `/api/telescope_states`, then I think the mocked message should be

```
'ConnectionError - Error retrieving telescope states'
```

mgdaily (approved):
Looks good, one minor adjustment

tom_observations/tests/facilities/test_ocs.py
mgdaily: https://github.com/observatorycontrolsystem/observation-portal/pull/272/files#diff-a3c10e98ffff710567f79c0422a5c60508a394efa56ec839f1f40fd7bcd64e7dR62

phycodurus (approved):

PR #688, opened 2023-09-19
Feature/open outside links in new tabs

jchate6:
This PR changes all of the external links I could find to open in a new tab so as to not derail a users TOM experience.

Specifically, these changes affect:

- Links to Alert Brokers
- Home Page Links
- Facility links for additional info when making observing requests
- Links to site weather information


jchate6:
@phycodurus can you think of any other repos that might need this update?

phycodurus (approved):

PR #694, opened 2023-10-02
Fix Issues with DataProduct Deletion in DeleteView 

davner:
This pull request addresses two issues related to the deletion of `DataProduct` objects. The first issue was that the files were not being properly deleted, causing name collisions for newly added files. This is in context of when adding previous observational data by obs ID.

In the previous version, deleting a `DataProduct` was being ignored because:
```
DeleteView uses FormMixin to handle POST requests. 
As a consequence, any custom deletion logic in DataProductDeleteView.delete() handler 
should be moved to form_valid().
```
This problem was noticed when a random string was being appended to filenames, due to collisions caused by the delete() method being ignored, after being deleted and re-added. Django appends such a string to avoid filename collision when the file is not actually deleted.

The second issue was that the `DataProduct` reference was not being removed, leading to errors in subsequent calls to creating data product context. This was only realized after moving to `form_valid()`.

Changes: 
- Moved the deletion logic from `DataProductDeleteView.delete()` to `form_valid()` to comply with Django's `DeleteView` behavior. This ensures the associated file is actually deleted.
- Added a call to `data_product.delete()` to remove the `DataProduct` reference from the database, resolving the `None` reference issue.

This image shows how the filename was changed after deleting and re-adding.
Pasted image 20231002085627

I am using Django 4.2.5

### DIFF
--- a/tom_dataproducts/views.py
+++ b/tom_dataproducts/views.py
@@ -190,17 +190,30 @@ class DataProductDeleteView(Raise403PermissionRequiredMixin, DeleteView):
         referer = urlparse(referer).path if referer else '/'
         return referer
 
-    def delete(self, request, *args, **kwargs):
+    def form_valid(self, form):
         """
-        Method that handles DELETE requests for this view. First deletes all ``ReducedDatum`` objects associated with
-        the ``DataProduct``, then deletes the ``DataProduct``.
+        Method that handles DELETE requests for this view. It performs the following actions in order:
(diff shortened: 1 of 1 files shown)


jchate6:
Straight from the [Django 4.0 Documentation](https://docs.djangoproject.com/en/4.2/releases/4.0/#generic-views)

jchate6 (approved):
Most beautiful PR I've seen in ages.
Thanks for this.

Issue #611, opened 2023-02-08
Add ATLAS catalog search app


rachel3834:
**Is your feature request related to a problem? Please describe.**
ATLAS is a ground-based optical imaging survey which provides public access to its forced photometry on its timeseries imaging.  
A user can submit on-sky coordinates and a timestamp and the service will perform photometry on any image matching those criteria on their archive, and return photometry data. 

Their data service offers an API which is documented here:
https://fallingstar-data.com/forcedphot/apiguide/

The user needs an account and a token to access the service. 

**Describe the solution you'd like**
As a TOM user, I would like to be able to query the ATLAS archive by sending a set of coordinates (a Target) and a timestamp.  If the archive's response indicates that they have a lightcurve available, I would like my TOM to ingest the resulting lightcurve as a ReducedDatum, and associate it with my Target.

PR #696, opened 2023-10-06
Feature/add forced photometry

jnation3406:
This adds Forced Photometry services on the "Manage Data" Tab of the target details page. The Atlas service is implemented here and will work synchronously or asynchronously if dramatiq is added to the project. There is a guide on setup and customization. 

I might want to change the query form to pop out in a modal later, but its not necessary for the functionality so I won't hold up the story for that.


jchate6:
Does this address Issue #611 ?

tom_dataproducts/data_processor.py
```diff
@@ -65,3 +66,10 @@ def process_data(self, data_product):
         :rtype: list of 2-tuples
         """
         return []
+
+    def data_type_override(self):
```
jchate6, line 70: What is the motivation for this?
Just to provide a tool for multi-type DataProducts?

phycodurus (approved):

tom_dataproducts/templates/tom_dataproducts/forced_photometry_form.html
```diff
@@ -0,0 +1,14 @@
...
+<link rel="stylesheet" href="{% static 'tom_common/css/main.css' %}">
+{% endblock %}
+{% block content %}
+{{ form|as_crispy_errors }}
+<h2>Query {{ form.service.value }} Forced Photometry Service</h2>
+<br />
+<h4>Target <b>{{ target.name }}</b> at RA {{ target.ra }}, DEC {{ target.dec }}</h4>
+<br />
+{% crispy query_form %}
+{% endblock %}
```
jchate6, line 14: Wants a blank line

tom_dataproducts/views.py
```diff
@@ -86,6 +87,86 @@ def post(self, request, *args, **kwargs):
         )
 
 
+class ForcedPhotometryQueryView(LoginRequiredMixin, FormView):
```
jchate6, line 90: Should we have a test for any of this with some mocked data tot est our processors, etc?

jchate6: We should also have some canary tests to make sure the brokers don't change but that might be it's own Issue/PR.

tom_dataproducts/tasks.py
```diff
@@ -0,0 +1,103 @@
...
+                print(f"{resp.status_code} {message}")
+                t_sec = re.findall(r"available in (\d+) seconds", message)
+                t_min = re.findall(r"available in (\d+) minutes", message)
+                if t_sec:
+                    waittime = int(t_sec[0])
+                elif t_min:
+                    waittime = int(t_min[0]) * 60
+                else:
+                    waittime = 10
+                print(f"Waiting {waittime} seconds")
```
jchate6, line 53: Do we want to leave these print statements in? Should these be logger messages?

tom_dataproducts/data_processor.py
jnation3406: I described it in the comment for that method... But basically its for if you want the ReducedDatum to have a different datatype then the DataProduct. I.e. atlas_photometry DataProduct makes photometry ReducedDatums so that they show up in the photometry plots.

tom_dataproducts/views.py
jnation3406: Yeah I wasn't sure how useful unit tests would be. The best test is actually hitting the service, and if the service goes down / changes its data structure then thatll break things anyway.

tom_dataproducts/data_processor.py
jchate6: Yeah, I was confused about how the Reduced Datums and Data Products data types were tied together. Lindy Walked me through it.
Realistically, There's no reason we should ever use the Data Product type for the Reduced Datums. That should always be part of the data processor's job. A Data Product shouldn't even have the type "Photometry" It should be "Photometry Table" or "Fits file" etc.

jnation3406: I agree with that, I didn't make the change to have it supplied by the data processor because that wouldn't have been backwards compatible with other data processors that people may have written already...

jchate6: understandable.

tom_dataproducts/views.py
jchate6: makes sense.

jchate6 (approved):

Issue #697, opened 2023-10-16
Tab pane not set in local storage for `/targets/1`


davner:
**Describe the bug**
The Jquery to store the tab pane in `target_detail.html` in `tom_targets/templates/tom_targets` does not function as expected. The pane is not remembered or reloaded on refresh. 

**To Reproduce**
1. Create a target
2. Go to `http://localhost:8000/targets/1/`
3. Click on a tab pane other than `Observe` such as `Observations`
4. Reload the page
5. See that the tab pane is not remembered
6. Adding a `console.log` to the javascript that does the storing to local storage to the tab event doesn't not get triggered.

**Expected behavior**
We expect the tab pane to be remembered (stored) and loaded to that pane upon refresh.

**Desktop (please complete the following information):**
- Chrome
- Dev branch commit `4f0729b`
- Python `3.10.13`

**Proposed Solution**
**Option 1 (preferred)**
Rather than storing the tab pane to local storage, have the ability to load a specific pane by adding in a url param `tab`. The vanilla javascript to accomplish this:
```javascript
document.addEventListener("DOMContentLoaded", function() {
  // Fetch the URL and look for the 'tab' query parameter.
  const url = new URL(window.location.href);
  const tabQuery = url.searchParams.get('tab');

  // If a tab is specified in the URL, make it active.
  if (tabQuery) {
    const activeTab = '#' + tabQuery;
    // Remove the 'tab' query parameter to avoid unwanted persistence across
    // page reloads.
    url.searchParams.delete('tab');
    history.replaceState({}, document.title, url.toString());

    // Show the active tab, if any.
    const tabElement = document.querySelector(`a[href="${activeTab}"]`);
    if (tabElement) {
      tabElement.click();
    }
  }
});
```
Then in the `views.py` for whatever TOM package, add it to the url to open to that pane. 
`return redirect(reverse('tom_targets:detail', args=(target_id,)) + '?tab=observations')`

This will redirect to the url `http://localhost:8000/targets/1/?tab=observations`, flips the tab pane to `Observations` and then strips the param and sets the url to `http://localhost:8000/targets/1/`. 

This gives us the advantage of being able to load any pane upon return such as using the buttons in `/targets/:
- `Update Observation Status`
- `Add Existing Observation`
- `Upload` for photometry

**Option 2**
Do the previous behavior of storing the tab pane viewed in local storage and restore it. 

[...]

PR #701, opened 2023-10-18
Add ability to navaigate to specific tab in target view.

davner:
Resolves Issue #697.

1. Introduces URL parameter support for directly navigating to specific tabs in the target detail view (e.g., `/targets/1/?tab=observations`). Initial implementation is limited to two sections that our users found most impactful but could be applied anywhere the directs to the target detail view. This was written with vanilla javascript, removing the concern of Jquery conflicts. 
  
2. Augments `observation_list.html` to display observation IDs, improving user experience to trace data to what ID in the overview.

### DIFF
--- a/tom_observations/templates/tom_observations/partials/observation_list.html
+++ b/tom_observations/templates/tom_observations/partials/observation_list.html
@@ -1,9 +1,10 @@
 <table class="table table-striped">
-  <thead><tr><th>Facility</th><th>Created</th><th>Status</th><th>Scheduled</th><th>Saved  data</th><th>View</th></tr></thead>
+  <thead><tr><th>Facility</th><th>Observation ID</th><th>Created</th><th>Status</th><th>Scheduled</th><th>Saved  data</th><th>View</th></tr></thead>
   <tbody>
   {% for observation in observations %}
     <tr>
       <td>{{ observation.facility }}</td>
+      <td>{{ observation.observation_id }}</td>
       <td>{{ observation.created }}</td>
       <td>{{ observation.status }}</td>
       <td>{{ observation.scheduled_start }}</td>
--- a/tom_observations/templates/tom_observations/partials/observation_list.html
+++ b/tom_observations/templates/tom_observations/partials/observation_list.html
@@ -4,7 +4,11 @@
   {% for observation in observations %}
     <tr>
       <td>{{ observation.facility }}</td>
-      <td>{{ observation.observation_id }}</td>
+      {% if observation.url %}
+        <td><a href="{{ observation.url }}">{{ observation.observation_id }}</a></td>
+      {% else %}
+        <td>{{ observation.observation_id }}</td>
+      {% endif %}
       <td>{{ observation.created }}</td>
       <td>{{ observation.status }}</td>
       <td>{{ observation.scheduled_start }}</td>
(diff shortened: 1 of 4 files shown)


tom_observations/templates/tom_observations/partials/observation_list.html
```diff
@@ -1,9 +1,10 @@
 <table class="table table-striped">
-  <thead><tr><th>Facility</th><th>Created</th><th>Status</th><th>Scheduled</th><th>Saved  data</th><th>View</th></tr></thead>
+  <thead><tr><th>Facility</th><th>Observation ID</th><th>Created</th><th>Status</th><th>Scheduled</th><th>Saved  data</th><th>View</th></tr></thead>
   <tbody>
   {% for observation in observations %}
     <tr>
       <td>{{ observation.facility }}</td>
+      <td>{{ observation.observation_id }}</td>
```
jchate6, line 7: What do you think of replacing this line with
```
{% if observation.url %}
  <td><a href="{{ observation.url }}">{{ observation.observation_id }}</a></td>
{% else %}
  <td>{{ observation.observation_id }}</td>
{% endif %}
```
So that this can be a direct link to the data source if it has one?

jchate6 (changes requested):
I really like these features, and I think this new functionality is fantastic, but I don't see how this directly addresses Issue #697. 
The active tab information is still lost on refresh, correct? 
Does this functionality override that issue, or should the issue persist once this PR is merged?

davner:
@jchate6 , you've raised a valid point. The problem still persists, as this PR provides enhancements, but doesn't address the original issue. I've given this some thought and have something. I do want to avoid local storage because I don't want the tab to persist upon navigating away. 

1. Event Listener on Tabs: I will add an event listener to the tabs.
2. URL Update on Tab Click: When a tab is clicked, the URL will be updated to reflect the active tab. This ensures the user stays on the same tab after certain actions.
3. Page Refresh Behavior: Refreshing the page will take the user back to the last active tab, as indicated by the URL.
4. No Local Storage Dependency: We won't use local storage, so the last opened tab won't persist if the user navigates away. This keeps the experience clean and straightforward.
5. Specific Tab Redirects: The solution still allows for redirecting to a specific tab when needed.

Our URLs would look like
http://localhost:8000/targets/1/?tab=manage-datahttp://localhost:8000/targets/1/?tab=manage-groups
etc when the user clicks on a tab. Then if refreshed the correct tab persists. 


Thoughts?

tom_observations/templates/tom_observations/partials/observation_list.html
davner: Yes, this is a great idea!

jchate6:
Avoiding local storage makes a lot of sense. 
I agree that an event listener on the tab is the way to go. 
Thanks!

davner:
@jchate6 Comments have been addressed and tab is remembered on reload.

jchate6:
Thanks @davner! I probably won't get to this until Monday.

jchate6 (approved):
This is great.
Thanks for this @davner.
I think we'll probably want to take some time in the future changing redirects and maybe adding a default option to `settings.py` so users can choose their default landing page for their TOM. Very exciting.

davner:
@jchate6 Not a problem. I'm trying to push back all the changes I make for our project that would be beneficial to all users. Glad you like this one!

PR #718, opened 2023-10-26
Feature/install poetry

phycodurus:
We should consider widening the requirements on the dependencies listed in the pyproject.toml file


pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+authors = ["William (Lindy) Lindstrom <llindstrom@lco.global>"]
+license = "GPLv3"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.9"
+numpy = "<2"
+astroquery = ">=0.4.2"
+astroplan = "~0.8"
+astropy = "^5.3.4"
```
jchate6, line 14: This is incompatible with python 3.8

jchate6 (changes requested):
I strongly disagree with how these have been done. I think we need as broad a range of dependencies as possible, especially if we plan to support currently supported versions of python. (3.8 has security support for another year).

Also, the original motivation for including poetry was specifically to allow us to increase our compatibility with more versions of astropy and numpy. This was a request from our largest user and should probably at least be considered in this PR.

pyproject.toml
```diff
@@ -0,0 +1,53 @@
+[tool.poetry]
+name = "tom-base"
+version = "2.15.0"
+description = "TOM Toolkit and base modules"
+authors = ["William (Lindy) Lindstrom <llindstrom@lco.global>"]
+license = "GPLv3"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
jchate6, line 10: If we are doing this, we should remove the tests for 3.8

pyproject.toml
jchate6: noted ;)

jchate6:
https://nsf-noirlab.gitlab.io/csdc/antares/client/installation.html:
```Note
Python 3.10 Prerequisites

librdkafka is required to install antares-client with python 3.10

Mac: brew install librdkafka

Ubuntu: apt-get install librdkafka-dev

**Comment by jchate6** [comment]
Library on github:
https://github.com/confluentinc/librdkafka

File: poetry.lock [hunk 3]
```diff
@@ -0,0 +1,2660 @@
...
+python-versions = ">=3.7"
+files = [
+    {file = "antares-client-1.4.0.tar.gz", hash = "sha256:8cc6e0166da68882da31d704e03b626f5e62d1f2b0896eca9235161a97664998"},
+]
+
+[package.dependencies]
+astropy = "*"
+bson = "*"
+click = "*"
+confluent_kafka = "1.7.0"
```
jchate6, line 28: This is the restriction that's incompatible with Python >3.9

poetry.lock
```diff
@@ -0,0 +1,2660 @@
...
+description = "Antares broker module for the TOM Toolkit"
+optional = false
+python-versions = "*"
+files = [
+    {file = "tom-antares-1.2.6.tar.gz", hash = "sha256:42706cf971ce3b33ea583f6de6d7edbea891f64dc31ef27c6bf9eb774bcb400e"},
+    {file = "tom_antares-1.2.6-py3-none-any.whl", hash = "sha256:f86fe971b18f66cdd687848ab1bcea1704eab278f953b1b29a5f0db01f7e5c13"},
+]
+
+[package.dependencies]
+antares-client = ">=1.4,<2.0"
```
jchate6, line 2425: This is the dependency that we include that leads to the 3.10 incompatability.

jchate6: antares_client requires a version of confluent_kafka that requires a c++ package to be compatible with python >3.9

jchate6:
`pip install tom-antares==1.2.6` Gets me the following:

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tom-alertstreams 0.5.1 requires hop-client<0.8,>=0.7, but you have hop-client 0.8.0 which is incompatible.
Successfully installed antares-client-1.4.0 confluent_kafka-1.7.0 tom-antares-1.2.6

.github/workflows/run-tests.yml
```diff
@@ -9,20 +9,21 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: "3.9" # for antares-client's confluent-kafka dependency
```
jchate6, line 12: This is technically now for numpy not playing nice with 3.12?

jchate6 (approved):
Very exciting to have this, Do we need to update documentation for not having tom_antares anymore?

Also, do we need to make an issue in tom_antares to handle the dependency fallout that will no doubt occur there?

.github/workflows/run-tests.yml
phycodurus: exactly

PR #734, opened 2023-11-01
Add user object into facility class so it can be used for per-user au…

jnation3406:
…thentication. This was a request by Bob Eldering to support a facility which authenticates per user against an OIDC server. Having the user within the facility class allows whatever facility operation you want to retrieve an OAuthToken stored for that user/facility and use that in requests to the facility. I think I've added the user in a backwards compatible way for our current facilities, and in all the calls that might want it for the facility. I've tested this locally with the LCO facility and it is still working.

### DIFF
--- a/tom_observations/facilities/soar.py
+++ b/tom_observations/facilities/soar.py
@@ -76,8 +76,8 @@ class SOARFacility(LCOFacility):
         'SPECTRA': SOARSpectroscopyObservationForm
     }
 
-    def __init__(self, user=None, facility_settings=SOARSettings('LCO')):
-        super().__init__(user=user, facility_settings=facility_settings)
+    def __init__(self, facility_settings=SOARSettings('LCO')):
+        super().__init__(facility_settings=facility_settings)
 
     def get_form(self, observation_type):
         return self.observation_forms.get(observation_type, SOARImagingObservationForm)
--- a/tom_observations/facilities/soar.py
+++ b/tom_observations/facilities/soar.py
@@ -76,8 +76,8 @@ class SOARFacility(LCOFacility):
         'SPECTRA': SOARSpectroscopyObservationForm
     }
 
-    def __init__(self, facility_settings=SOARSettings('LCO')):
-        super().__init__(facility_settings=facility_settings)
+    def __init__(self, user=None, facility_settings=SOARSettings('LCO')):
+        super().__init__(user=user, facility_settings=facility_settings)
 
     def get_form(self, observation_type):
         return self.observation_forms.get(observation_type, SOARImagingObservationForm)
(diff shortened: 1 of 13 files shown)


jchate6 (changes requested):
Can we get this feature mentioned in the [docs](https://tom-toolkit.readthedocs.io/en/latest/observing/observation_module.html)? Maybe just a short comment that this exists and how to use it near the bottom?

jchate6:
Also, would it make sense to write a test in `tom_observations.tests` tracing this variable through to make sure we don't lose it in future changes to `facility.py`.

tom_observations/facilities/soar.py
```diff
@@ -76,8 +76,8 @@ class SOARFacility(LCOFacility):
         'SPECTRA': SOARSpectroscopyObservationForm
     }
 
-    def __init__(self, facility_settings=SOARSettings('LCO')):
-        super().__init__(facility_settings=facility_settings)
+    def __init__(self, user=None, facility_settings=SOARSettings('LCO')):
+        super().__init__(user=user, facility_settings=facility_settings)
```
jchate6, line 80: I don't think this is backwards compatible with existing facility modules. Wouldn't they all need this change as well?

jchate6 (changes requested):

tom_observations/facilities/soar.py
jnation3406: Do you know any other modules based on the OCS module? It is backwards compatible as long as the other module does not override __init__, and if it does override __init__ and calls super then it is still backwards compatible but the user will just be None... The only case where it wouldn't be is if they subclassed the OCSFacility and overroad the __init__ basing in facility_settings as the first param and not a kwarg. I don't think there are any other OCS facility subclasses other than what we have provided though...

jchate6: It's not just OCS facilities though. you changed BaseObservationFacility. so I think every facility that isn't an OCS facility will break without the above code with something like `xxxFacility.__init__() got an unexpected keyword argument 'user'`

jnation3406: yeah I guess you are write if they have overridden the __init__ method in their facility. Do you know of other facilities outside tom_base? The only one I see in the tomtoolkit github is tom_swift that would need to be updated...

jchate6: We have tom_swift and tom_lt, and griffin has one for the mmt, and I know we had several people experimenting with this sort of thing at the workshop, but ultimately we don't know (or I at least don't know) if other people have made their own custom facility modules.

jnation3406: Okay... I guess I can change it to not be initialized. It will just add an extra line of code everywhere its initialized.

jnation3406: I've changed it to set the user after instantiation, so this should be backwards compatible now.

jchate6 (approved):
Looks good!
Thanks!

PR #743, opened 2023-11-14
715 pan starrs forced photometry

### LINKED ISSUE #715: Pan-STARRS forced photometry
Integrate new forced photometry service into tom_base

phycodurus:
Add PanSTARRS PS1 forced photometry service class, MAST client library (panstarrs_api.py), and data processor.

This is a draft while I add some error checking, but comments are welcome.


jchate6 (changes requested):
All of this needs to be much more explicitly defined in the documentation.
[This section](https://tom-toolkit.readthedocs.io/en/stable/managing_data/forced_photometry.html#integrating-existing-forced-photometry-services) has code blocks that do not show up correctly in the live docs. 

I'd also like to see a list of currently built in Forced Photometry services with details about each specific one including links to the API and description of features.

Finally, I think `Query Forced Photometry Service` section on the Target Detail page's Manage Data Tab should be under the data table and either needs to be hidden if no services are available, or should include a blurb below it saying something like "Forced Photometry has not been initialized. Please see the [documentation](https://tom-toolkit.readthedocs.io/en/latest/managing_data/forced_photometry.html) for details." *Edit: I've added the "no forced photometry warning" to PR #761

tom_dataproducts/forced_photometry/forced_photometry_service.py
```diff
@@ -8,6 +8,7 @@
 from django.utils.module_loading import import_string
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
jnation3406, line 11: I assume you mean to remove this...

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,235 @@
...
+class PanstarrsForcedPhotometryQueryForm(fps.BaseForcedPhotometryQueryForm):
+    # TODO: if we include both calendar and mjd fields, they should be linked in the UI
+    # min_date = forms.CharField(
+    #     label='Min date:', required=False,
+    #     widget=forms.TextInput(attrs={'class': 'ml-2', 'type': 'datetime-local'})
+    # )
+    # max_date = forms.CharField(
+    #     label='Max date:', required=False,
+    #     widget=forms.TextInput(attrs={'class': 'ml-2', 'type': 'datetime-local'})
+    # )
```
jnation3406, line 32: Did you want to do this the same way as the atlas form, where its an either/or for a datetimepicker or raw mjd input? You should be able to copy that code from the atlas form...

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,235 @@
...
+
+    min_detections = forms.IntegerField(
+        label='Minimum detections:',
+        initial=2, required=False,
+        help_text=('Objects with nDetections=1 tend to be artifacts, so this is a way to'
+                   ' eliminate most spurious objects from the catalog.')
+    )
+
+    data_release = forms.ChoiceField(
+        # TODO: get these from panstarrs_api.py
```
jnation3406, line 50: You already did this TODO right?

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,235 @@
...
+        label='Data release: ',
+        choices=get_data_release_choices(),
+        initial='dr2',
+    )
+
+    # TODO: get these from panstarrs_api.py
```
jnation3406, line 56: and this one as well.

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
jnation3406: Oh i see you prefill in the mjd so you wouldn't know which input the user intends to use I guess without some more code.

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,235 @@
...
+        cleaned_data = super().clean()
+        logger.debug(f"PanstarrsForcedPhotometryQueryForm.clean() -- cleaned_data: {cleaned_data}")
+
+        # TODO: update cross-field validation
+        if not (cleaned_data.get('min_date') or cleaned_data.get('min_date_mjd')):
+            raise forms.ValidationError("Must supply a minimum date in either datetime or mjd format")
+        # if cleaned_data.get('min_date') and cleaned_data.get('min_date_mjd'):
+        #     raise forms.ValidationError("Please specify the minimum date in either datetime or mjd format")
+        # if cleaned_data.get('max_date') and cleaned_data.get('max_date_mjd'):
+            raise forms.ValidationError("Please specify the maximum date in either datetime or mjd format")
```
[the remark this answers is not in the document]
jnation3406: Yeah, this was for the atlas form that had two options for inputting date, to make sure that only one option was used for each field.

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,235 @@
...
+
+        target = Target.objects.get(pk=query_parameters.get('target_id'))
+
+        # first, convert datetime fields to mjd if necessary
+        min_date_mjd = query_parameters.get('min_date_mjd')
+        if not min_date_mjd:
+            min_date_mjd = Time(query_parameters.get('min_date')).mjd
+        max_date_mjd = query_parameters.get('max_date_mjd')
+        if not max_date_mjd and query_parameters.get('max_date'):
+            max_date_mjd = Time(query_parameters.get('max_date')).mjd
```
jnation3406, line 142: Since you aren't using a min_date/max_date field, this code should be removed.

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,235 @@
...
+            'epochMean.max': max_date_mjd,
+            'nDetections.min': query_parameters.get('min_detections'),
+        }
+
+        catalog = query_parameters.get('catalog')
+        data_release = query_parameters.get('data_release')
+        response = mast_query(table=catalog, release=data_release, request_data=request_data)
+
+        # Did the response contain any data?
+        if len(response.content) == 0:
```
jnation3406, line 183: Unless this always responds nearly instantly (which I assume it doesn't for large date ranges), then you should probably still do this in the same exact way as the Atlas query, making it a task that is callable directly if no dramatiq is setup, or calling it asynchronously if dramatiq is present. This way, if the user has dramatiq setup, they will not have to sit waiting (and potentially navigate away breaking things) when making a large query. Instead, it will return back to them immediately and they can continue working until the query is done and data is available.

tom_dataproducts/templates/tom_dataproducts/partials/dataproduct_list_for_target.html
```diff
@@ -23,7 +23,7 @@ <h4>Data</h4>
           {% include 'tom_dataproducts/partials/js9_button.html' with url=product.data.url only %}
         {% endif %}
       </td>
-      <td><a href="{{ product.data.url }}">{{ product.get_file_name }}</a></td>
+      <td><a href="{{ product.data.url }}" target="_blank">{{ product.get_file_name }}</a></td>
```
jnation3406, line 26: what is this needed for?

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs_api.py
```diff
@@ -0,0 +1,168 @@
...
+        ('stack', 'Stacked object'),
+        ('forced', 'Forced mean object'),
+        ('detection', 'Detections'),
+    ]
+
+    if data_release == 'dr1':
+        choices = [  # DR1 choices
+            ('mean', 'Mean object'),
+            ('stack', 'Stacked object'),
+        ]
```
jnation3406, line 52: this could just be an if/else setting choices either way?

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs_api.py
```diff
@@ -0,0 +1,168 @@
+import logging
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+import requests
+
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
jnation3406, line 10: Don't forget to remove this

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def get_base_url():
+    """return the PanSTARRS MAST API base URL from settings
+    """
+    try:
+        base_url = settings.FORCED_PHOTOMETRY_SERVICES.get(
+            'PANSTARRS', {}).get('url', 'https://catalogs.mast.stsci.edu/api/v0.1/panstarrs/')
```
jnation3406, line 18: Should the default url just be in the settings.py? and all this code just to return your fixed value doesn't seem great. If the user doesn't have a PANSTARRS section then they shouldn't be able to get a url right? I would think they should just not be able to submit a panstarrs query in that case...

jnation3406 (changes requested):
Some comments, biggest change being you probably should use the task setup for the main logic of querying panstarrs so it can happen asynchronously for the user (even though it is a single synchronous query).

tom_dataproducts/templates/tom_dataproducts/partials/dataproduct_list_for_target.html
phycodurus: I thought the href-ed data product should open in a new browser tab.

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
phycodurus: #848

tom_dataproducts/templates/tom_dataproducts/forced_photometry_form.html
```diff
@@ -7,8 +7,13 @@
...
 {{ form|as_crispy_errors }}
 <h2>Query {{ form.service.value }} Forced Photometry Service</h2>
+{% if info_url %}
+<p><em>More information: <a href="{{ info_url }}" target="_blank"> {{ form.service.value }}</a></em></p>
+{% endif %}
 <br />
-<h4>Target <b>{{ target.name }}</b> at RA {{ target.ra }}, DEC {{ target.dec }}</h4>
-<br />
+<h4>Target <b><a href={% url "targets:detail" target.id %}> {{ target.name }}</a></b>
+    at RA {{ target.ra|floatformat:4 }}, DEC {{ target.dec|floatformat:4 }}</h4>
```
jchate6, line 15: If it isn't going to be a form field, we need to include radius here. Preferably in arcmin or arcsec.

jchate6 (changes requested):
We should handle an undefined DATA Product type more gracefully.
Currently the code crashes if an unknown dataproduct type is referenced from the target detail page:
`tom_base/tom_dataproducts/views.py", line 158`
`tom_base/tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py", line 189`

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if data_release == 'dr1':
+        choices = [  # DR1 choices
+            ('mean', 'Mean object'),
+            ('stack', 'Stacked object'),
+        ]
+    return choices
+
+
+def get_default_columns(catalog='mean'):
```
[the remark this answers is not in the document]
phycodurus: for the moment, we'll be limiting queies to the Mean object catalog

tom_dataproducts/processors/panstarrs_processor.py
```diff
@@ -0,0 +1,88 @@
...
+            for datum in data:
+                # extract the timestamp from the epochMean column
+                time = Time(float(datum['epochMean']), format='mjd')
+                utc = TimezoneInfo(utc_offset=0*units.hour)
+                time.format = 'datetime'
+                timestamp = time.to_datetime(timezone=utc)
+                for optical_filter in ['g', 'r', 'i', 'z', 'y']:
+                    # these filter and column names come from pastarrs_api.py
+                    mag_col_name = f'{optical_filter}MeanPSFMag'
+                    mag_err_col_name = f'{optical_filter}MeanPSFMagErr'
```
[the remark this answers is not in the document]
phycodurus: At the moment, we will only support queries to the Mean object catalog. The 'Stacke object' catalog is not yet available through this form.

tom_dataproducts/templates/tom_dataproducts/forced_photometry_form.html
phycodurus: see Issue #880

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,232 @@
...
+import tom_dataproducts.forced_photometry.forced_photometry_service as fps
+from tom_dataproducts.models import DataProduct
+from tom_dataproducts.exceptions import InvalidFileFormatException
+from tom_dataproducts.data_processor import run_data_processor
+
+from .panstarrs_api import get_data_release_choices, get_catalog_choices, mast_query
+
+logger = logging.getLogger(__name__)
+# logger.setLevel(logging.DEBUG)
+
```
jchate6, line 21: need an extra blank line here

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,227 @@
...
+        submits the query to the service
+
+        Called from views.py ForcedPhotometryQueryView.post() if form.is_valid()
+        """
+        logger.debug(f"Querying PanSTARRS service with params: {query_parameters}")
+
+        target = Target.objects.get(pk=query_parameters.get('target_id'))
+
+        # first, convert datetime fields to mjd if necessary
+        min_date_mjd = query_parameters.get('min_date_mjd')
```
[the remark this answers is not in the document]
phycodurus: #881

tom_dataproducts/processors/panstarrs_processor.py
```diff
@@ -0,0 +1,88 @@
...
+from astropy.time import Time, TimezoneInfo
+
+from tom_dataproducts.data_processor import DataProcessor
+from tom_dataproducts.exceptions import InvalidFileFormatException
+
+logger = logging.getLogger(__name__)
+# logger.setLevel(logging.DEBUG)
+
+
+class PanstarrsProcessor(DataProcessor):
```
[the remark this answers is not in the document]
jchate6: This is caused by different csv files for the Data Products. So this is also true if I pull DR1 vs DR2

tom_dataproducts/forced_photometry/panstarrs_service/panstarrs.py
```diff
@@ -0,0 +1,227 @@
...
+        min_time = Time(min_date_mjd, format='mjd').datetime.strftime('%Y_%m_%d')
+        dp_name = f"panstarrs-{data_release}-{catalog}_{min_time}"
+        if max_date_mjd:
+            max_time = Time(max_date_mjd, format='mjd').datetime.strftime('%Y_%m_%d')
+            dp_name += f"-{max_time}"
+        dp_name += '.csv'
+
+        file = ContentFile(response.content, name=dp_name)
+
+        dp, created = DataProduct.objects.get_or_create(
```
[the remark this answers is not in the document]
jchate6: I'll also point out that this breaks when trying to run forced photometry for different targets with the same date range.

jchate6 (approved):

PR #746, opened 2023-11-16
Add Observation Deletion Functionality.

davner:
- Implement unified method for deleting observations and associated data products.
- Resolve bug preventing deletion of thumbnails.
- Introduce new template for observation deletion process.

This PR introduces a new feature for deleting observations and their associated data products, responding to user feedback within our project. It also addresses a oversight made by my previous PR that prevented the deletion of thumbnails for a dataproduct, ensuring a more comprehensive cleanup process.

A new template has been added to streamline the observation deletion process. Tests are added and confirm that observations and related data products are thoroughly deleted. Additionally, a deletion button has been integrated into the observation detail view for user convenience.

I believe these enhancements could benefit the `TOMToolkit` community. However, I understand and respect if this functionality does not align with current priorities.


dmcollom:
I'm very far outside of the TOM Toolkit community at this point, so hopefully this isn't too out of the blue, but it seems to me that it might be more desirable to have a bulk data delete as a script, but for an individual delete to simply mark an observation as "deleted" with a `deleted_at` timestamp, and then hiding those from any views.

To me, the use cases for deletion are to remove visual clutter, and to improve performance (which feels more in line with data product deletion), but there may still be value in keeping a record of the observation, even if the data that came out of it wasn't of value.

Obviously up to the maintainers, just saw this come through and figured I'd offer my two cents.

davner:
I agree that a way to bulk delete data is important, and that's a request from our users as well. I will be implementing that soon. I see it being beneficial to offer both capabilities. The function I wrote that loops through an observation record could be used as the bulk deleter.

griffin-h:
Sorry to jump in here, but we were just discussing a related issue in our group. Is there a way to cancel an observation through the TOM? I'm not aware of one, but it would be helpful. This would not delete the `ObservationRequest`, but rather send a signal to the observatory API to cancel the request, and set the `ObservationRecord`'s status to `'CANCELED'`. Let me know if I should open a separate issue on this.

rachel3834:
You're very welcome to comment here @Hosseinzadeh, Griffin - (griffin0)
***@***.***> !

It may be a separate issue, since both functions would be very useful for a
lot of TOM users.

griffin-h:
> It may be a separate issue, since both functions would be very useful for a lot of TOM users.

Great, I made a separate issue, #786.

jchate6 (commented):
I'm not sure I understand everything you are going for here. 
Specifically, the code in `tom_common.utils` seems strange to me.

It seems like all of this work should be done in the Model?
It's not clear to me why adding the `data_product.data.delete()` and the `data_product.thumbnail.delete()` to an extended `DataProduct.delete()` in the model wouldn't accomplish all of this through cascades.

Is there something I'm missing?

PR #763, opened 2023-12-01
only show unknown status warning if there are any

griffin-h:
This is a very small change that hides the yellow "observations with unknown status" warning at the top of every target page unless there is actually an observation request with unknown status. Currently, I find myself ignoring that message because it appears at the top of every page, so I hope this will make it more useful. This makes it match the behavior of the green "upcoming observations" message, which only appears if there are pending observations.


jchate6 (approved):
Thanks for this!

PR #769, opened 2023-12-06
Add Override for data Product path so user can define whatever path they like

jchate6:
This should allow a user to add a `DATA_PRODUCT_PATH` value to their `settings.py` allowing for them to override the default path creation method for Data Products.

### DIFF
--- a/docs/managing_data/customizing_data_processing.rst
+++ b/docs/managing_data/customizing_data_processing.rst
@@ -57,11 +57,10 @@ you wanted to set the `data_product_path` to ``{target}/{facility}/{observation_
 .. code:: python
 
    def custom_data_product_path(data_product, filename):
-       return '{}/{}/{}/{}'.format(data_product.target.name,
-                                   data_product.observation_record.facility,
-                                   data_product.observation_record.observation_id,
-                                   filename)
-
+       return f'{data_product.target.name}/' \
+               f'{data_product.observation_record.facility}/' \
+               f'{data_product.observation_record.observation_id}/' \
+               f'{filename}'
 
 All data products are automatically “processed” on upload, as well. Of
 course, that can mean different things to different TOMs! The TOM has
--- a/docs/managing_data/customizing_data_processing.rst
+++ b/docs/managing_data/customizing_data_processing.rst
@@ -48,6 +48,21 @@ In order to add new data product types, simply add a new key/value pair,
 with the value being a 2-tuple. The first tuple item is the database
 value, and the second is the display value.
 
+When a ``DataProduct`` is uploaded, its path is set by a `data_product_path`. By default this is set to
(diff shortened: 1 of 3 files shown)


docs/managing_data/customizing_data_processing.rst
```diff
@@ -48,6 +48,21 @@ In order to add new data product types, simply add a new key/value pair,
...
+When a ``DataProduct`` is uploaded, its path is set by a `data_product_path`. By default this is set to
+``{target}/{facility}/{filename}``, but can be overridden by setting the ``DATA_PRODUCT_PATH`` in
+``settings.py`` to a dot-separated method path that takes a ``DataProduct`` and filename as arguments. For example, if
+you wanted to set the `data_product_path` to ``{target}/{facility}/{observation_id}/{filename}``, you would point the
+``DATA_PRODUCT_PATH`` to a custom method that looks something like this:
+
+.. code:: python
+
+   def custom_data_product_path(data_product, filename):
+       return '{}/{}/{}/{}'.format(data_product.target.name,
```
phycodurus, line 60: Let's use an f-string here.

tom_dataproducts/models.py
```diff
@@ -85,11 +93,21 @@ def data_product_path(instance, filename):
...
-    else:
-        return '{0}/none/{1}'.format(instance.target.name, filename)
+    try:
+        path_class = settings.DATA_PRODUCT_PATH
+        try:
+            mod_name, class_name = path_class.rsplit('.', 1)
+            mod = import_module(mod_name)
+            clazz = getattr(mod, class_name)
+        except (ImportError, AttributeError):
+            raise ImportError('Could not import {}. Did you provide the correct path?'.format(path_class))
```
phycodurus, line 103: let's use an f-string here as well.

tom_dataproducts/models.py
```diff
@@ -85,11 +93,21 @@ def data_product_path(instance, filename):
...
+        return clazz(instance, filename)
+    except AttributeError:
+        # Uploads go to MEDIA_ROOT
+        if instance.observation_record is not None:
+            return '{0}/{1}/{2}'.format(instance.target.name, instance.observation_record.facility, filename)
+        else:
+            return '{0}/none/{1}'.format(instance.target.name, filename)
```
phycodurus, line 110: more f-string opportunities :-)

phycodurus (approved):

davner:
This looks extremely promising, thank you! I can get around to testing it later today or first thing tomorrow.

docs/managing_data/customizing_data_processing.rst
jchate6: the line breaks are so much uglier... :(

PR #770, opened 2023-12-06
Fix/uncredentialed facility pages

jchate6:
This PR provides some default values and behaviors to the OCS facility that allow OCS facility observation forms to be displayed but not submitted when the facility is not yet configured in `settings.py`.


tom_observations/facilities/ocs.py
```diff
@@ -963,6 +988,16 @@ def __init__(self, *args, **kwargs):
         if isinstance(self, CadenceForm):
             self.helper.layout.insert(2, self.cadence_layout())
 
+    def button_layout(self):
+        target_id = self.initial.get('target_id')
```
phycodurus, line 992: maybe  add a comment or two to leave bread crumbs:
 * what method, where in the class hierarchy?, is being overridden here
 * how is the super class method being extended or modified 
 * and why

tom_observations/views.py
```diff
@@ -227,6 +227,11 @@ def get_context_data(self, **kwargs):
         facility_context = facility.get_facility_context_data(target=target)
         context.update(facility_context)
 
+        try:
+            context['unconfigured'] = ", ".join(facility.facility_settings.check_configuration())
```
phycodurus, line 231: It looks like the value of `context['unconfigured'] can be either a Boolean or a <str> and that's a little sus to me.
At the very least, there should be some context here for what this key:value is, how and where, it's used. etc.

tom_observations/facilities/ocs.py
```diff
@@ -81,6 +97,10 @@ def __init__(self, facility_name):
     def get_setting(self, key):
         return settings.FACILITIES.get(self.facility_name, self.default_settings).get(key, self.default_settings[key])
 
+    def check_configuration(self):
+        """ Check that the settings for this facility are present, and return list of unconfigured settings"""
```
phycodurus, line 101: not sure what a "list of unconfigured settings" means. maybe provide an example to clarify?

phycodurus (approved):

tom_observations/views.py
jchate6: This relies on the falsy nature of an empty string. I've changed it to only return a string.

Issue #767, opened 2023-12-05
Review of new TOM creation process and documentation


rachel3834:
The purpose of this issue is to review the creation of a basic TOM following the documentation, to identify any broken features or missing documentation.  Items that have been checked off have been tested and found to work. 

For reference, this testing was done on a Macbook Pro running OS/X Sonoma, using Python 3.11 in a new virtual environment. 

### Documentation
- [x] **PRIORITY** Getting started doc should have a more obvious button at the top of the TOM's readthedocs page and/or a 'Quick start' link in the navbar. 
- [x] Github README should list development team and add a section on how to cite/acknowledge the Toolkit 
- [x] Readthedocs should have a clear section on how to cite the Toolkit (inc plugins crediting contributors)
- [x] Getting started guide indicates python >=3.7 is required - is this still current?  Do we still support 3.6?
- [x] Getting started guide: "./manage.py tom_setup.py" will only work if the user has cd'd into the project's top level directory first. This step should be explained. 
- [x] Getting started guide procedure works cleanly.  But the user can't do much in the TOM without logging in - this step should be described at the end. 
- [x] Documentation could use a new page describing Targets, and how the TOM handles different parameters.  For example, the page on [TOM Targets](https://tom-toolkit.readthedocs.io/en/stable/targets/index.html#) gives useful detailed references but no overall overview of what a Target is, and the parameters that are stored (i.e. the attributes on the Target model) and what are not - and why it was designed this way.  This leads into a discussion of the role of Tags and Extra Parameters and what the difference is. 
- [x] Documentation needs a description of Target Groups and how to use them. 
- [ ] **PRIORITY** Documentation needs a high-level page describing the overall principle of how observatories interact with the TOM, and how the authentication and proposal information is configured in the settings.  Then it needs to link to pages on how to configure each of the default supported observatories, and the observing modes that they offer.  Gemini in particular since the template system (and how to create them) isn't obvious.  
- [x] Documentation on LCO's observing modes, and how to use the TOM's forms to configure them.  For example, the observing request form offers me "configurations" and "instrument configurations" - what's the difference?
- [x] What is the difference between "observations" and "observation groups" under the TOM's top navbar options for the 
 "Observations" menu?  Not mentioned in the documentation. 
- [x] Add documentation section to describe TOM Users and User Groups.  This should include a description of what the AnonymousUser is for. 

 
#### Target Import
- [X] Manual creation of non-sidereal target works
[...]

PR #775, opened 2023-12-07
Docs/unittests for toms

rachel3834:
Added documentation on how to implement unittests, for those users who are customizing their TOM systems.  
This addresses one item from Issue #767 .

### DIFF
--- a/docs/customization/testing_toms.rst
+++ b/docs/customization/testing_toms.rst
@@ -4,7 +4,10 @@ Testing TOMs
 As functionality is added to a TOM system it is highly beneficial to write
 `unittests <https://docs.python.org/3/library/unittest.html>`__ for each
 new function and module added.  This allows you to easily and repeatably
-test that the function behaves as expected, and makes error diagnosis faster.
+test that the function behaves as expected, and makes error diagnosis much faster.
+Although it can seem like extra work, ultimately writing unittests will save
+you time in the long run.  For more on why unittests are valuable,
+`read this <https://docs.djangoproject.com/en/5.0/intro/tutorial05/>`__.
 
 The TOM Toolkit provides a built-in framework for writing unittests,
 including some TOM-specific factory functions that enable you to
@@ -14,11 +17,25 @@ functionality.  This allows test code to be written which acts on a
 stand-alone test database, meaning that any input used for testing
 doesn't interfere with an operational TOM database.
 
-Code Structure
---------------
+Code Structure and Running Tests
+--------------------------------
+There are two main options on where the test code lives in your TOM,
+depending on how you want to run the tests.  Here we follow convention and refer to
+the top-level directory of your TOM as the `project` and the subdirectory
+for the TOM as the `application`.  Although they are often called the same
+name, the distinction matters because TOMs can have multiple `applications`
+within the same `project`.  The actual test code will be the same regardless
+of which option you use - this is described in the next section.
 
-It's good practise to place your unittest code in a single subdirectory
-in the top-level directory of your TOM, normally called ``tests/``:
+**Option 1: Use the built-in manage.py application to run the tests**
+
+With this option, test code is run using ``manage.py``, just like management
+commands.  ``manage.py`` is designed to consider any file with a filename
+prefixed by ``test_`` within the project directory and subdirectories as test code,
+and it will find them automatically.
+
+If you chose this option, then it's good practise to place test code in the subdirectory for
+the application that they apply to, in a file called ``tests.py``.    The code structure for this option would be:
 
 ::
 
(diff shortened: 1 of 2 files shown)


docs/customization/testing_toms.rst
```diff
@@ -0,0 +1,311 @@
...
+--------------------------------
+There are two main options on where the test code lives in your TOM,
+depending on how you want to run the tests.  Here we follow convention and refer to
+the top-level directory of your TOM as the `project` and the subdirectory
+for the TOM as the `application`.  Although they are often called the same
+name, the distinction matters because TOMs can have multiple `applications`
+within the same `project`.  The actual test code will be the same regardless
+of which option you use - this is described in the next section.
+
+**Option 1: Use the built-in manage.py application to run the tests**
```
jchate6, line 30: If you use the above strategy for subheadings, you could make this a sub-subheading by adding yet another section style (`=====`).

docs/customization/testing_toms.rst
```diff
@@ -0,0 +1,311 @@
...
+   ├── static.
+   ├── templates/
+   └── tmp/
+
+In order to test this, we need to set up some input data in the form of
+coordinates.  We could do this just by setting some input RA, Dec values
+as purely numerical attributes.  However, bearing in
+mind that the TOM stores this information as entry in its
+database, a more realistic test would present that information in the
+form of a `Target object <targets/index>`__.  The Toolkit includes a number of
```
jchate6, line 246: Relative links like this don't work the same as external links.

````Target object <targets/index>`__````
should be 
```
:doc:`Target object <../targets/index>`
```

File: docs/customization/testing_toms.rst [hunk 3]
```diff
@@ -0,0 +1,311 @@
...
+``factory`` classes designed to make it easy to create realistic input
+data for testing purposes.  The last element that the ``setUp`` method needs
```
jchate6, line 248: Minor point, but I'm not sure I agree with this. It's often better to define the expected output in the test method since the specific outcomes will often depend on what test you are actually conducting. (For instance the `calc_gal_coords` could have have a sexigesimal or decimal degree flag and you might want to test both use cases with the same setup).

jchate6: Maybe change this to something less definitive than "needs"?
"One last element that is often useful in the ``setup`` method..."

jchate6 (changes requested):
What a fantastic summary of TOM Testing!
Thanks for this!

docs/customization/testing_toms.rst
```diff
@@ -0,0 +1,311 @@
...
+The TOM Toolkit provides a built-in framework for writing unittests,
+including some TOM-specific factory functions that enable you to
+generate examples of TOM database entries for testing.  This is, in turn,
+built on top of `Django's unittest framework <https://docs.djangoproject.com/en/5.0/topics/testing/overview/>`__, inheriting a wealth of
+functionality.  This allows test code to be written which acts on a
+stand-alone test database, meaning that any input used for testing
+doesn't interfere with an operational TOM database.
+
+Code Structure and Running Tests
+--------------------------------
```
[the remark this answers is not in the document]
rachel3834: Ah thank you, that's the helpful reference I was looking for.  I've used ~~~~ for subsections.

docs/customization/testing_toms.rst
rachel3834: Implemented.  I used ++++ for subsubheadings because PyCharm apparently doesn't like '====' for some reason.

docs/customization/testing_toms.rst
rachel3834: Updated

docs/customization/testing_toms.rst
rachel3834: That's fair.  Different frameworks take a different approach to this.  In pytest for instance you are expected to declare both in a decorator for each test method.  I think that level of detail goes beyond this documentation, so I have moved the expected values to the test method and removed this "One last element..." line from the description of the setUp method to avoid confusion.

docs/customization/testing_toms.rst
```diff
@@ -0,0 +1,313 @@
...
+   ├── static.
+   ├── templates/
+   └── tmp/
+
+In order to test this, we need to set up some input data in the form of
+coordinates.  We could do this just by setting some input RA, Dec values
+as purely numerical attributes.  However, bearing in
+mind that the TOM stores this information as entry in its
+database, a more realistic test would present that information in the
+form of a `Target object <../targets/index>`.  The Toolkit includes a number of
```
jchate6, line 249: you need a `:doc:` before this for the link to work.

Should be 
```
:doc:`Target object <../targets/index>`
```

jchate6 (changes requested):
One last thing...

jchate6 (approved):
Amazing!

PR #776, opened 2023-12-08
Correct error loading observation template forms

jchate6:
This PR will allow for Gemini templates to have the same behavior as OCS facility template pages and load a stub rather than a 500 error when loading an unconfigured LT template.


tom_observations/views.py
```diff
@@ -611,6 +611,16 @@ class ObservationTemplateCreateView(FormView):
     def get_facility_name(self):
         return self.kwargs['facility']
 
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+
```
phycodurus, line 616: let's compare how `tom_swift` is supplying a custom Template through the Facility and this method. Then we can standardize on the one we like or undstand why each method is best for it's use case.

PR #778, opened 2023-12-08
Fix Incorrect Time Format in `DATETIME_FORMAT` Setting

### LINKED ISSUE #638: Black Hole Tom is Displaying Datetimes incorrectly
[Black Hole TOM](https://github.com/BHTOM-Team/bhtom2) has reported an issue with a discrepancy between the displayed Datetime and the stored datetime for the creation date of a target

The following conversation with @wyrzykow has been copied form the TomToolkit slack channel:

> I have added an extra field (creation_date) as datetime type. It writes the correct UTC datetime after the creation of a target, also the correct time is when displaying Update form. It is correct also in the database (seen as admin). However, in target page, target_detail (target_data), the time is modified. See below:
image
image

> is possible django is parsing the datetime incorrectly? Or maybe I am writing my datetime incorrectly? This is the database entry:
image

> I have just created a target at 7:59 UT (my extra field has: 2023-03-07T07:59:12.570020+00:00), however, target.created returns 7:03 UT, also in the tom/admin list:

davner:
This PR updates the `DATETIME_FORMAT` setting to correct a bug that caused the wrong time to be displayed in our application.

Previously, the DATETIME_FORMAT was set as:

```python
DATETIME_FORMAT = 'Y-m-d H:m:s'
```

However, this format string incorrectly used `m` (month) in the place where `i` (minute) should have been used. 

The corrected format string is:

```python
DATETIME_FORMAT = 'Y-m-d H:i:s'
```

This should resolve #638. 🙂

### DIFF
--- a/tom_setup/templates/tom_setup/settings.tmpl
+++ b/tom_setup/templates/tom_setup/settings.tmpl
@@ -151,7 +151,7 @@ USE_L10N = False
 
 USE_TZ = True
 
-DATETIME_FORMAT = 'Y-m-d H:m:s'
+DATETIME_FORMAT = 'Y-m-d H:i:s'
 DATE_FORMAT = 'Y-m-d'
 
 


jchate6 (approved):
Good Catch.
This is a bit insidious. Users who have already downloaded their tom will need to change this in their `settings.py` if they haven't already.

davner:
True. I made a PR in the black hole TOM group with the change to make them aware.

jchate6:
Thanks. I'll try to alert other users.

davner:
Also, at least this only affects the display and not the logged datetime.

jchate6:
Documentation for Reference:
https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#std:templatefilter-date

PR #785, opened 2023-12-14
remove extra space from text areas in Firefox

griffin-h:
There is a known issue that adds extra space to Bootstrap text areas in Firefox: https://github.com/twbs/bootstrap/issues/23295

This adds a workaround so that single-line text areas are the same height as char fields in all browsers. Example screenshots attached. Note the height of the first field.

Screenshot 2023-12-14 at 14-11-33 SAGUARO Report to TNS
Screenshot 2023-12-14 at 14-36-11 SAGUARO Report to TNS

### DIFF
--- a/tom_common/static/tom_common/css/main.css
+++ b/tom_common/static/tom_common/css/main.css
@@ -21,3 +21,9 @@ body {
 #login-buttons > li {
   margin: 0 2px 0 2px;
 }
+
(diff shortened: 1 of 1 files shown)


phycodurus (approved):
We couldn't reproduce this, but this change doesn't break anything (AFAWCT).
Thanks, Griffin, for all your help! See you soon :-)

PR #789, opened 2023-12-18
Enhancement: Extend settings to set default plotly theme.

davner:
This PR introduces the ability for users to customize the `plotly` theme directly from `settings.py`. By default, the theme is set to `plotly_white`, maintaining the current default behavior. 

This change empowers users with more flexibility in customizing the appearance of their Plotly charts, aligning with their personal or organizational branding and styling preferences. In the future, this should be extended to allow users to provide their own templates.

### DIFF
--- a/tom_common/apps.py
+++ b/tom_common/apps.py
@@ -1,5 +1,18 @@
 from django.apps import AppConfig
+from django.conf import settings
+import plotly.io as pio
 
 
 class TomCommonConfig(AppConfig):
     name = 'tom_common'
+
+    def ready(self):
+        # Set default plotly theme on startup
+        valid_themes = ['plotly', 'plotly_white', 'plotly_dark', 'ggplot2', 'seaborn', 'simple_white', 'none']
+
+        # Get the theme from settings, default to "plotly_white".
+        plotly_theme = getattr(settings, 'PLOTLY_THEME', 'plotly_white')
+        if plotly_theme not in valid_themes:
+            plotly_theme = 'plotly_white'
+
+        pio.templates.default = plotly_theme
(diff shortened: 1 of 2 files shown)


tom_setup/templates/tom_setup/settings.tmpl
```diff
@@ -322,6 +322,10 @@ REST_FRAMEWORK = {
     'PAGE_SIZE': 100
 }
 
+# Default Plotly theme setting, can set to any valid theme:
```
jchate6, line 325: We should add a mention of this functionality in the docs.
I'm thinking in https://tom-toolkit.readthedocs.io/en/stable/managing_data/plotting_data.html would make the most sense for now.

In the future, we will likely want to clean up the settings.tmpl and move all of these non-critical customization options into the docs.

jchate6: see Issue #793

tom_common/apps.py
```diff
@@ -1,5 +1,18 @@
 from django.apps import AppConfig
+from django.conf import settings
+import plotly.io as pio
 
 
 class TomCommonConfig(AppConfig):
     name = 'tom_common'
+
+    def ready(self):
```
jchate6, line 9: Might be a good idea to call the super here in case that's ever needed.

jchate6 (approved):

jchate6:
@davner I haven't looked into it yet, but I can't merge dev into this branch like I can for your other PRs from your personal repo.
I'm guessing `gemini-hlsw` has different permissions?

davner:
@jchate6 that is strange, because PR #787 was from `gemini-hlsw`.

jchate6:
> @jchate6 Do you want me to make the requested changes or will you handle it?

I'll handle it! Those are mostly notes for me.

jchate6:
> @jchate6 that is strange, because PR #787 was from `gemini-hlsw`.

Yeah, I just merged that one manually on my local machine.

davner:
Good to now. I'll figure out the permissions or just switch back to my personal repo.

Issue #816, opened 2024-01-16
Redirect to specific tabs on Target Detail page


jchate6:
Now that we can, we may want to go directly to certain tabs on the target detail page.
Redirects exist in `tom_base` in the following locations:

- [ ] tom_dataproducts/tests/tests.py:238
- [ ] tom_dataproducts/tests/tests.py:244
- [ ] tom_dataproducts/tests/tests.py:305
- [ ] tom_dataproducts/tests/tests.py:321
- [ ] tom_dataproducts/tests/tests.py:69
- [ ] tom_dataproducts/tests/tests.py:205
- [ ] tom_dataproducts/tests/tests.py:211
- [x] tom_dataproducts/templates/tom_dataproducts/dataproduct_list.html:31 --> Manage_data
- [x] tom_dataproducts/templatetags/dataproduct_extras.py:116  -->  Manage_data
- [x] tom_dataproducts/views.py:158  -->  Photometry
- [x] tom_dataproducts/views.py:374  --> Default
- [x] tom_dataproducts/views.py:422 --> Default
- [x] tom_observations/facilities/lt.py:17 --> observe
- [x] tom_observations/facilities/ocs.py:1003 --> observe
- [x] tom_observations/templates/tom_observations/observation_list.html:48 --> observations
- [x] tom_observations/facility.py:93 --> Observe
- [x] tom_observations/views.py:367 --> Observations
- [x] tom_observations/views.py:428 --> Observations
- [x] tom_observations/views.py:475 --> Observations
- [x] tom_observations/templates/tom_observations/observation_form.html:10 --> Default 
- [x] tom_targets/templates/tom_targets/partials/recent_targets.html:7 --> Default
- [x] tom_targets/templates/tom_targets/partials/recently_updated_targets.html:7 --> Default
- [ ] tom_targets/templates/tom_targets/partials/target_plan.html:3 --> possibly broken by change to Default
- [x] tom_targets/templates/tom_targets/partials/target_table.html:32 --> Default
- [x] tom_targets/templates/tom_targets/target_form.html:33 --> Default
- [x] tom_targets/templates/tom_targets/target_detail.html:93 --> Reloads to current tab.
- [x] tom_targets/templates/tom_targets/target_detail.html:123 --> Default
- [x] tom_targets/templates/tom_targets/target_share.html:10 --> Default
- [x] tom_targets/templates/tom_targets/target_share.html:33 --> Default
- [ ] tom_targets/tests/tests.py:89
[...]

Issue #745, opened 2023-11-16
Add integration points to templates so additional modules can be added without overriding templates


jchate6:
The Basic Idea here is to implement "Integration points" that would behave in a similar way to the observation facilities, with additional modules automatically being linked into reasonable locations without a TOM admin having to override a template.
Essentially this would allow a user to install a new TOM module and have links and buttons for the module automatically appear in the proper places.

This would reduce user maintenance and allow users to more easily benefit from future improvements to the TOM toolkit.

Obvious Integration points:

- [x] Target Detail Buttons
- [ ] Target Detail Tabs
- [ ] Nav Bar

phycodurus:
For a general overview of the strategy, please see PR #941 (especially commits:
 1. f1974f31564dc9fcf99cd6021692e032688f5b58
 2. 00d0cb2835adaae8b353d63a4a3711cdbcfb9a44
 3. e1d7fc9bf0f2cb4dcd257a3dfde0d0e2a0480903
These commits show the basic idea which is to define a `TOMToolkitAppConfig` that TOMToolkit INSTALLED_APPs can inherit from (it's a subclass of django.app.AppConfig). `tom_base` and go
to the Django App registry and introspect the apps there looking for the extensions that
`TOMToolkitAppConfig` (and it's INSTALLED_APPs subclasses) define.

jchate6:
#973 Should include information about these integration points.

PR #819, opened 2024-01-18
update redirects to target_detail page

jchate6:
Small PR to enhance redirects to the target detail page so the user arrives at the tab that makes the most sense (to me.)
Many instances have been left to the "default" which will eventually be selectable by the individual TOM creator.

See Linked issue #816 for changed redirects.


phycodurus (approved):
See additional comments in Issue #745

PR #840, opened 2024-02-12
Added view to enable target selection for manual observing runs

rachel3834:
This feature aims to improve the TOM's useability for observers operating more traditional, block-scheduled observing facilities, often manually or through remote operations.  The feature provides a form and view to enable them to easily identify targets visible at their facility on the date indicated.


tom_targets/views.py
```diff
@@ -749,3 +752,97 @@ def post(self, request, *args, **kwargs):
...
+        ] + settings.SELECTION_EXTRA_FIELDS
+        # for param in settings.SELECTION_EXTRA_FIELDS:
+        #    table_columns.append(param)
+
+        # Calculate the visibility of all selected targets on the date given
+        # Since some observatories include multiple sites, the visibiliy_data returned is always
+        # a dictionary indexed by site code.  Our purpose here is to verify whether each target is ever
+        # visible at lower airmass than the limit from any site - if so the target is considered to be visible
+        observable_targets = []
+        for object in targets:
```
jchate6, line 820: This should probably be `for target in targets:`
`object` is a [python built-in](https://docs.python.org/3/library/functions.html#object)

tom_targets/views.py
```diff
@@ -749,3 +752,97 @@ def post(self, request, *args, **kwargs):
...
+        # The displayed table can be extended to include selected extra_fields for each target,
+        # if configured in the TOM's settings.py. So we set the list of table columns accordingly.
+        table_columns = [
+            'Target', 'RA', 'Dec', 'Site', 'Min airmass'
+        ] + settings.SELECTION_EXTRA_FIELDS
+        # for param in settings.SELECTION_EXTRA_FIELDS:
+        #    table_columns.append(param)
+
+        # Calculate the visibility of all selected targets on the date given
+        # Since some observatories include multiple sites, the visibiliy_data returned is always
```
jchate6, line 816: *visibility

tom_targets/views.py
```diff
@@ -749,3 +752,97 @@ def post(self, request, *args, **kwargs):
...
+                object, start_time, end_time,
+                visibiliy_intervals, airmass_max,
+                observation_facility=request.POST.get('observatory')
+            )
+            for site, vis_data in visibility_data.items():
+                airmass_data = np.array([x for x in vis_data[1] if x])
+                if len(airmass_data) > 0:
+                    s = SkyCoord(object.ra, object.dec, frame='icrs', unit=(u.deg, u.deg))
+                    target_data = [
+                        object.name, s.ra.to_string(u.hour), s.dec.to_string(u.deg, alwayssign=True),
```
jchate6, line 833: `object.name` should be a link back to the target's observe page in case the user wants to actually submit an observation.

tom_targets/forms.py
```diff
@@ -203,3 +204,18 @@ class TargetListShareForm(forms.Form):
...
         self.fields['share_destination'].choices = get_sharing_destination_options()
+
+
+class TargetSelectionForm(forms.Form):
+    """
+    Form for selecting the targets from a pre-existing TargetList
+    """
+    target_list = forms.ModelChoiceField(
+        TargetList.objects.all(),
+        required=False)
```
jchate6, line 215: Not requiring a `TargetList` means that if the user doesn't select anything, then the code will try to do this for EVERY target. This will likely completely crash any TOM with a large DB as it tries to get visibilities for a few million objects.

jchate6 (changes requested):
I've made some specific comments, that I think are important, but I have some general  comments that I think will require a larger discussion.

Specifically, I'd like to talk about what you are actually looking to accomplish here. This returns a selection of targets in a list that have an airmass < 2.0 within a specific 24 hour window with 10 steps starting at YYYY-MM-DDT00:00:00 for a list of telescopes at a specific observatory. That feels like a highly specific use case that should probably be made more generalized. Users might want to change the limiting airmass, or see all of the telescopes at once, or set up their window so it encompasses a full night at CPT rather than splitting it.

Additionally, I feel like the information returned is limiting. This doesn't tell a user WHEN the target will be at it's lowest airmass, or sort them in any way or give a user an idea of how long the target will be up. Also, this is completely incompatible with non-sidereal targets. 

Ultimately I'm wondering if this whole idea makes more sense being incorporated into the target list view. An airmass filter could possibly be added, and "observation planning" columns could be put on the table.

Would love to discuss further when you have time.

tom_targets/views.py
rachel3834: Agreed, done.

rachel3834:
I've implemented all of the changes requested above.  I take your point about the potential for more generalized functionality for this view but I think it makes sense to implement this version first before trying to generalize.

tom_targets/templates/tom_targets/target_facility_selection.html
```diff
@@ -0,0 +1,52 @@
...
+            {% bootstrap_field form.target_list %}
+        </div>
+        <div class="col-sm-1">
+            <input type="submit" class="btn btn-primary" formaction="{% url 'targets:target-selection' %}" id="submit_target_selection" value="Submit" name="target_selection_form" style="position:absolute; bottom:1rem"">
+        </div>
+    </div>
+  </form>
+</div>
+<div class="row">
+    {% if observable_targets %}
```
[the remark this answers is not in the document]
rachel3834: When I implemented this is, the displayed page had an empty table and displayed the no-targets-found message all of the time, before the form had been submitted.  I found this a bit counter intuitive so I've added a test for the table_columns, which is only available when the form has been filled in and submitted.  In that case, if no targets are found, a table is displayed with a no-targets entry.  Otherwise the target table appears.

rachel3834:
I've refactored the FormView to implement pagination on the list of targets, and updated the get and post methods so that pagination works as intended in the template.

tom_targets/views.py
```diff
@@ -750,14 +755,13 @@ def post(self, request, *args, **kwargs):
...
 
-class TargetGroupingView(PermissionListMixin, ListView):
+class TargetGroupingView(PermissionListMixin, FormView):
     """
     View that handles the display of ``TargetList`` objects, also known as target groups. Requires authorization.
     """
     permission_required = 'tom_targets.view_targetlist'
     template_name = 'tom_targets/target_grouping.html'
     model = TargetList
-    paginate_by = 25
```
jchate6: We also require something in `paginate_by` since we call `{% bootstrap_pagination page_obj extra=request.GET.urlencode %}` in the template.

jchate6 (changes requested):
I've pointed out a few broken or problematic changes. I haven't really gotten back into the actual utility of the feature. It seems like there is some pagination issue, but I can't really tell if that's just a symptom of other problems.

tom_targets/forms.py
```diff
@@ -282,3 +286,24 @@ def __init__(self, *args, **kwargs):
...
+
+class TargetSelectionForm(forms.Form):
+    """
+    Form for selecting the targets from a pre-existing TargetList
+    """
+    target_list = forms.ModelChoiceField(
+        TargetList.objects.all(),
+        required=True)
+    facilities = [(x, x) for x in get_service_classes()]
+    facilities += [(x.full_name, x.full_name) for x in ObservingFacility.objects.all()]
```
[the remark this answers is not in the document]
rachel3834: Thanks for explaining.  I've added a function to tom_observations.utils to build the combined list of facilities and moved the choice field configuration to the __init__ method.

tom_targets/views.py
```diff
@@ -750,14 +755,13 @@ def post(self, request, *args, **kwargs):
         return redirect(reverse('tom_targets:list') + '?' + query_string)
 
 
-class TargetGroupingView(PermissionListMixin, ListView):
```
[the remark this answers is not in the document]
rachel3834: This class has a FormView rather than a ListView.

tom_observations/utils.py
```diff
@@ -34,10 +42,14 @@ def get_sidereal_visibility(target, start_time, end_time, interval, airmass_limi
     :param airmass_limit: maximum acceptable airmass for the resulting calculations
     :type airmass_limit: int
 
-    :param observation_facility: observing facility for which to calculate the airmass. None indicates all available
-        facilities.
+    :param observation_facility: observing facility declared as a facility class for which to calculate the airmass.
```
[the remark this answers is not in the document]
rachel3834: OK.  I've refactored this so that the resolution of which observatory is requested is resolved in the get_sidereal_visibility function, and converted this parameter to a string.  I checked the other calls to this function also.

tom_targets/views.py
```diff
@@ -907,6 +910,159 @@ def post(self, request, *args, **kwargs):
...
+
+        context['target_visibilities'] = target_visibilities
+        context['targets_page'] = targets_page
+        context['use_table'] = True
+
+        return render(request, self.template_name, context)
+
+    def get(self, request, *args, **kwargs):
+
+        context = self.get_context_data(*args, **kwargs)
```
jchate6: add `request` to the arguments here too.

tom_targets/views.py
```diff
@@ -907,6 +910,159 @@ def post(self, request, *args, **kwargs):
...
+        visibiliy_intervals = 10
+
+        # Calculate the visibility of all selected targets on the date given
+        # Since some observatories include multiple sites, the visibility_data returned is always
+        # a dictionary indexed by site code.  Our purpose here is to verify whether each target is ever
+        # visible at lower airmass than the limit from any site - if so the target is considered to be visible
+        observable_targets = []
+
+        for target in targets_page:
+            start_time = datetime.combine(date, datetime.min.time())
```
jchate6: This means that a date will always be evaluated from 00UTC to 24UTC on the date in question. This could lead to unexpected visibilities depending on the local time of both the observer and the site.

For example, by default the value will be the current date, but if I ask for visibilities on 11/20 at 23:30 UTC I'll get YESTERDAY's visibilities, not visibility for TONIGHT.

I think it would make sense to either explicitly state that this is what you are doing in the help-text of the form, or to change this so start time is noon in the observatory's local time.

docs/observing/selecting_targets_for_facility.rst
```diff
@@ -0,0 +1,220 @@
...
+  :alt: Default table output for target selection
+
+Customizing the Selected Targets Table
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, this table will include the essential parameters necessary to point a
+telescope at the target, but it can be easily extended to add further information.
+
+The columns of the table can be configured by editing the TOM's ``settings.py`` file.
+Additional parameters can be defined for each target by adding dictionary definitions
```
jchate6: This seems confusing to me. I think we should separate the `EXTRA_FIELDS` conversation from the `SELECTION_EXTRA_FIELDS` description.

Instead, let's link them to `docs/targets/target_fields.rst` where they can learn how to add other fields, and then tell them how to use `SELECTION_EXTRA_FIELDS`

tom_targets/views.py
```diff
@@ -907,6 +910,159 @@ def post(self, request, *args, **kwargs):
...
+                        target.name,
+                        s.ra.to_string(u.hour),
+                        s.dec.to_string(u.deg, alwayssign=True),
+                        site,
+                        round(airmass_data.min(), 1)
+                    ]
+
+                    # Extract any requested extra parameters for this object, if available
+                    for param in getattr(settings, 'SELECTION_EXTRA_FIELDS', []):
+                        if param in target.extra_fields.keys():
```
jchate6: This should also look for fields in the Target Model itself, not just target extras.
People will be adding custom Target models for exactly this purpose.

docs/observing/selecting_targets_for_facility.rst
```diff
@@ -0,0 +1,220 @@
...
+In this example, we have added ``EXTRA_FIELDS`` named ``Mag_now``, ``Priority1``
+and ``Priority2``, which the user can set either by editing each Target's parameters
+directly, or programmatically.   Having done so, we can add those ``EXTRA_FIELDS`` to
+the target selection table by adding the parameter names to the list of ``SELECTION_EXTRA_FIELDS``.
+This produces the table displayed below.
+
+.. image:: target_selection_table_extra_fields.png
+  :alt: Target Selection table with additional columns added
+
+.. code:: python
```
jchate6: Is this section sill relevant? At the very least it feels non-sequitur

jchate6: This should probably all be included in or a reference to `docs/observing/observation_module.rst`

jchate6 (changes requested):
Sorry to go another round, but I think this is looking much better!
Let me know if you have any questions.

tom_targets/views.py
rachel3834: I also realise that this will be important for moving objects, but support for those is (regrettably!) beyond the scope of this issue.  I think a moving object visibility calculator would be a great addition to the Toolkit in general, but I'd like to develop that with input from our Solar System user groups.

tom_targets/templates/tom_targets/target_facility_selection.html
```diff
@@ -0,0 +1,64 @@
...
+    {% endif%}
+    <label id="displaySelected"></label>
+    <table class="table table-striped">
+        {%if use_table %}
+            <tr>
+                {% for col in table_columns %}
+                    <th>{{col}}</th>
+                {% endfor %}
+            </tr>
+            {% if target_visibilities|length > 0 %}
```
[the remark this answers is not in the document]
rachel3834: I've modified the table columns as discussed to remove the RA, Dec and to add more information about the rise/set times and time of minimum airmass instead.

tom_targets/views.py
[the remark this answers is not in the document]
rachel3834: Corrected back to FormView.  Not sure why that was changed.

tom_targets/templates/tom_targets/target_facility_selection.html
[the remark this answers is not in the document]
jchate6: I think we missed something here. This is not resolved.
Let's talk about this one.

docs/observing/selecting_targets_for_facility.rst
```diff
@@ -0,0 +1,76 @@
...
+The columns of the table can be configured by editing the TOM's ``settings.py`` file,
+and adding a list called ``SELECTION_EXTRA_FIELDS``, as shown in the example below.  The entries
+in this list can be one or more of the exact names of attributes on your custom Target model.
+The list can be left empty if no additional fields should be added to the table.
+
+.. code-block:: python
+
+    # settings.py
+    SELECTION_EXTRA_FIELDS = [
+        'Mag_now',
```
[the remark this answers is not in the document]
rachel3834: Agreed, done.

tom_targets/views.py
```diff
@@ -907,6 +910,165 @@ def post(self, request, *args, **kwargs):
...
+                        round(airmass_data[:, 1].min(), 1),
+                        airmass_data[:, 0][0].time().strftime("%H:%M:%S"),
+                        airmass_data[:, 0][imin].time().strftime("%H:%M:%S"),
+                        airmass_data[:, 0][-1].time().strftime("%H:%M:%S"),
+                        str(duration).split('.')[0]
+                    ]
+
+                    # Extract any requested extra parameters for this object, if available
+                    for param in getattr(settings, 'SELECTION_EXTRA_FIELDS', []):
+                        target_data.append(getattr(target, param, None))
```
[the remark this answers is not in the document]
rachel3834: Yes, I realise it doesn't yet give the user the ability to control how the column names appear in the table.  For the time being, I've explained this more fully in the documentation.

tom_targets/templates/tom_targets/target_facility_selection.html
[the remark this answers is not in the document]
rachel3834: To resolve the issue of large targetlists having some blank pages in the table (due to none of the targets on that page being visible), and still retain the perfomance of calculating visibilities for only a few targets at one, I've added null table row entries for any target that isn't visible, since this information can be useful for observers.

tom_targets/views.py
```diff
@@ -1061,10 +1061,24 @@ def get_observable_targets(self, targets_page, window_start, observatory):
...
-                        target_data.append(getattr(target, param, None))
-                    observable_targets.append(target_data)
+                # Populate the results table with null entries for any targets
+                # that are not visible
+                else:
+                    target_data = [
+                        target.id,
+                        target.name,
+                        site,
+                        '<'+str(airmass_max),
```
jchate6: I think this is should be a `>` here, so the airmass becomes >2...

jchate6 (approved):
One tiny change!

PR #842, opened 2024-02-13
Enhancement to the TNS broker to fetch classification

KKruszynska:
While developing the Microlensing Observing Platform, our customized version of the TOM Toolkit, I realized that it would be beneficial to grab a name and class from TNS (we use it mostly for classification). 

I think this might be useful for other TOM Toolkit users, but there are two issues with it that I am not sure how to solve:
-  the way the function runs now creates a TNS time-out when I run it for a large number of sources,
- I created a unit test for the new function, but it requires a TNS bot API key (in my local set-up, I use MOP's bot TNS API key). 

I would be happy to receive feedback on how to improve the new functionality to make it usable for other teams.


jchate6 (changes requested):
We can't have normal unit tests that directly hit the outside world.
The purpose of these unit tests is to test the internal TOM Toolkit code, and we don't want to spam external services when running normal tests.

We have Canary tests that are run less frequently that hit the external services to make sure everything is behaving as expected with those connections.

For this purpose, You'll want to override settings with a fake API token, and mock the response from TNS.

See `TestLasairBrokerClass` in `tom_alerts.tests.brokers.test_lasair` for an example.
You can also use the `responses` package to spoof specific external calls.
See `test_share_dataproduct_no_valid_responses` at `tom_dataproducts.tests.test:568` for an example of how that works.

KKruszynska:
Thank you, I will adjust the unit test accordingly.

jchate6:
I'll take a more in-depth look at this soon to see if there is some way to use the existing code or a minor re-factor of the existing code to accomplish your purpose here.

tom_alerts/brokers/tns.py
```diff
@@ -179,3 +179,53 @@ def to_generic_alert(cls, alert):
...
         )
+
+    @classmethod
+    def fetch_tns_name(cls, parameters):
+        '''
+        Modified version of fetch_alert from original TOM Toolkit.
+        Fetches a TNS objname for a target at requested coordinates and radius.
+        '''
+
+        data = {
```
jchate6, line 190: I don't think this is the place for this code. 
What I think would be better is to pull out the relavent parts of the above code into one or 2 additional functions.
Use these new functions to return the most useful product and then access and parse these results in your own custom Code for your TOM.

i.e., create a `fetch_tns_transients` that includes lines 126-140 above and is plugged into the appropriate locations and also `get_tns_object_info` that includes lines 143-152 above.

jchate6 (changes requested):
Let me know if this makes sense, or if you need any help to implement this.
You might also want to take a look at `tom_catalog.harvesters.tns.TNSHarvester` to see if this already does some of what you want.

tom_alerts/brokers/tns.py
KKruszynska: Hi Joey, I changed the code to reflect that. I tested it and it seems to work, but I commented the unit test because it requires a response from the TNS server.

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +191,102 @@ def to_generic_alert(cls, alert):
             mag=alert['discoverymag'],
             score=alert['name_prefix'] == 'SN'
         )
+
+    # @classmethod
```
jchate6, line 195: You should clean up these commented sections.

tom_alerts/brokers/tns.py
jchate6: nice. Now move `fetch_tns_transients` above `fetch_alerts` and replace lines 131->152 with the line 
```
transients = cls.fetch_tns_transients(parameters)
```

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +191,102 @@ def to_generic_alert(cls, alert):
...
+            )
+        }
+        response = requests.post(TNS_SEARCH_URL, data, headers=cls.tns_headers())
+        response.raise_for_status()
+        transients = response.json()
+
+        return transients
+
+    @classmethod
+    def get_tns_object_info(cls, parameters):
```
jchate6, line 278: The same exact thing you did above for pulling out `fetch_tns_transients` from `fetch alerts`, I'd like you to do again with this code.

tom_alerts/tests/brokers/test_tns.py
```diff
@@ -0,0 +1,71 @@
...
+            {
+                'objname': '2023xrs',
+                'prefix': 'SN',
+                'objid': 140367
+            }
+        ]
+    }
+}
+
+# class TestClassSN(TestCase):
```
jchate6, line 27: I agree that this test isn't super useful. I think we should probably remove this whole file until we write some dedicated canary tests.

If you want, it would be helpful to mock the response (like you did above) and make sure we are handling the alerts correctly.

jchate6 (changes requested):

tom_alerts/tests/brokers/test_tns.py
KKruszynska: I wanted to but fetch_alerts, detch_tns_transients and get_object_info are making queries to TNS and I have no way of intercepting the signal and putting in a mock response. I will remove this file.

jchate6: You can intercept those calls with the `responses` package.
https://github.com/getsentry/responses

jchate6: but it's OK to ignore that for now.

KKruszynska: Inside the function?

jchate6: Yes. Take a look [here](https://github.com/TOMToolkit/tom_base/blob/90adda925cdd2d6d8e1c6480be4dc5b486cc8c9c/tom_dataproducts/tests/tests.py#L567) for an example of how we use it in a different part of tomtoolkit.

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +163,52 @@ def to_generic_alert(cls, alert):
             mag=alert['discoverymag'],
             score=alert['name_prefix'] == 'SN'
         )
+
+    @classmethod
+    def fetch_tns_transients(cls, parameters):
```
jchate6, line 168: Can you add a docstring here describing what is expected for parameters?

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +163,52 @@ def to_generic_alert(cls, alert):
...
+        if parameters['days_ago'] is not None:
```
jchate6, line 169: `days_ago` and `min_date` might also not exist in parameters, similar to the values below.
If these are enforced, you might want to use a `try...except` to pass back some kind of error.

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +163,52 @@ def to_generic_alert(cls, alert):
...
+            )
+        }
+        response = requests.post(TNS_SEARCH_URL, data, headers=cls.tns_headers())
+        response.raise_for_status()
+        transients = response.json()
+
+        return transients
+
+    @classmethod
+    def get_tns_object_info(cls, parameters):
```
jchate6, line 200: A similar doc string here. 
Remember Parameters here doesn't have to be the same as above, but it does need to contain a key called "objname".

jchate6 (changes requested):

tom_alerts/brokers/tns.py
KKruszynska: done!

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +163,75 @@ def to_generic_alert(cls, alert):
             mag=alert['discoverymag'],
             score=alert['name_prefix'] == 'SN'
         )
+
+    @classmethod
+    def fetch_tns_transients(cls, parameters):
+    """
```
jchate6, line 169: Something weird with the indent here...

tom_alerts/brokers/tns.py
```diff
@@ -191,3 +163,75 @@ def to_generic_alert(cls, alert):
...
         )
+
+    @classmethod
+    def fetch_tns_transients(cls, parameters):
+        """
+        Args:
+            parameters: dictionary containing days_ago (str), min_date (str)
+            and either:
+                - Right Ascention, declination (can be deg, deg or h:m:s, d:m:s) of the target,
+                and search radius and search radius unit (str), or
```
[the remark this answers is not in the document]
jchate6: looks like these can be either `"arcsec"`, `"arcmin"`, or `"deg"`

jchate6 (approved):
Looks great, just one little doc comment.

tom_alerts/brokers/tns.py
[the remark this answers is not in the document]
KKruszynska: added

jchate6 (approved):
Looks Great! Thanks for all of the work!

PR #874, opened 2024-03-14
allow for DATA_PRODUCT_TYPES to be empty

jchate6:
This change allows for users to not include any `DATA_PRODUCT_TYPES` in their `settings.py`.
If none are found, then spectroscopy and photometry types are added.

The tests will break if the settings are actually blank because they reference settings directly, but this is not actually diagnostic of the code itself as much as the tests, and we actually want to test instances where `DATA_PRODUCT_TYPES` are set.


phycodurus (approved):
more breadcrumbs will be appreciated by our six-months from now selves.

tom_dataproducts/models.py
```diff
@@ -211,7 +219,7 @@ def save(self, *args, **kwargs):
         Saves the current `DataProduct` instance. Before saving, validates the `data_product_type` against those
         specified in `settings.py`.
         """
-        for _, dp_values in settings.DATA_PRODUCT_TYPES.items():
+        for dp_values in DATA_TYPE_OPTIONS:
```
phycodurus: by saying `for dp_type_value, dp_type_name in DATA_TYPE_THINGS` the `dp_values[0]` would become `dp_type_value` which makes explicit what the first element of the tuple actually is.

tom_dataproducts/models.py
```diff
@@ -225,7 +233,8 @@ def get_type_display(self):
         :returns: Display value for a given data_product_type.
         :rtype: str
         """
-        return settings.DATA_PRODUCT_TYPES[self.data_product_type][1]
+        data_product_type_dict = {k: v for k, v in DATA_TYPE_OPTIONS}
+        return data_product_type_dict[self.data_product_type][1]
```
phycodurus, line 239: again, if indexing into the tuple can be made more explicit as to what that value actually is, then that reduces the cognitive load on the reader.

tom_dataproducts/models.py
```diff
@@ -368,7 +377,8 @@ class Meta:
         get_latest_by = ('timestamp',)
 
     def save(self, *args, **kwargs):
-        for _, dp_values in settings.DATA_PRODUCT_TYPES.items():
+        # Validate data_type based on options in settings.py or default types
+        for dp_values in DATA_TYPE_OPTIONS:
```
phycodurus: ditto. see comments above.

phycodurus (changes requested):

tom_dataproducts/models.py
phycodurus: yes, makes sense to use `_` for the unused value, but I don't think `dp_values` describes what that loop-variable represents. Something like `dp_choice_text` or `dp_type_choice_text` would be more descriptive.

tom_base/settings.py
```diff
@@ -222,7 +222,9 @@
 #
 # tom_dataproducts configuration
 #
-# Define the valid data product types for your TOM. Be careful when removing items, as previously valid types will no
+# Define the valid data product types for your TOM.
+# This is a dictionary of tuples, with the first element being the type name and the second being the display name.
```
phycodurus, line 226: thank you. I think referencing `Field.choices` make sense and explains _why_ the first and second elements of the tuple are what they are.

tom_setup/templates/tom_setup/settings.tmpl
```diff
@@ -221,7 +221,9 @@ FACILITIES = {
     },
 }
 
-# Define the valid data product types for your TOM. Be careful when removing items, as previously valid types will no
+# Define the valid data product types for your TOM.
+# This is a dictionary of tuples, with the first element being the type name and the second being the display name.
```
phycodurus, line 225: see comment above wrt `Field.choices`

phycodurus (approved):

tom_dataproducts/models.py
jchate6: I'm going with `dp_type` to mimic other usage

PR #896, opened 2024-04-15
quotes around button title text

griffin-h:
This is a minor fix to allow mouseover (title) text for custom buttons.


jchate6 (approved):
Nice

PR #897, opened 2024-04-15
update get_moon to get_body

griffin-h:
This is a minor fix to avoid the deprecation warning for `get_moon` (see [here](https://docs.astropy.org/en/stable/api/astropy.coordinates.get_moon.html)) by replacing it with `get_body`.


griffin-h:
Resolves #890

jchate6 (approved):
Thanks for this!

PR #910, opened 2024-04-23
903 improve target match manager

jchate6:
This PR extends and improves the Target match manager to make it more general and more easily modified by users.


jchate6:
Things to do:

- [ ] Check that new target manager doesn't prevent updates to pre-existing targets that are not unique
- [ ] make sure cone search is using sidereal target
- [ ] come up with names @phycodurus likes

tom_targets/base_models.py
```diff
@@ -41,29 +43,137 @@
...
+
+        :param target: The target object to be checked against.
+
+        :return: queryset containing matching Target(s).
+
+        """
+        queryset = self.match_name(target.name)
+        return queryset
+
+    def match_name(self, name):
```
jchate6, line 88: "name" and "target" are already properties of `Target.matches`

phycodurus (approved):

Issue #609, opened 2023-02-07
Make Data validation and control class for Reduced Datums


jchate6:
ReducedDatum.values is unvalidated. We could provide structure and consistency to this field with a validation class that takes data and provides a uniform JSON response.

PR #911, opened 2024-04-25
rename magnitude_error to error

### LINKED ISSUE #790: Inconsistent usage of "error" or "magnitude_error" in ReducedDatum.value
**Describe the bug**
It appears that some parts of the code use "error" and others use "magnitude_error" for the uncertainty on the magnitude of a photometry point. For example, the photometry plot uses "error," but the photometry table uses "magnitude_error."

**To Reproduce**
1. Go to a target detail page with photometry
2. Click on the photometry tab
3. See some points with error bars plotted in the figure
4. Find those points in the table and see that they do not have errors listed

**Expected behavior**
I believe "error" was the original keyword used in the toolkit, because that is what's in the plot and what is parsed from photometry uploads. So I'd suggest replacing "magnitude_error" with "error" in `atlas_processor.py`, `alertstreams/hermes.py` and `dataproduct_extras.py` (and anywhere else I missed).

griffin-h:
This resolves #790 by making the naming convention for uncertainties on magnitudes consistent as "error". This should be resolved more robustly according to #609, but in the mean time, this allows the photometry table on the target detail page to work.

### DIFF
--- a/tom_dataproducts/alertstreams/hermes.py
+++ b/tom_dataproducts/alertstreams/hermes.py
@@ -162,8 +162,8 @@ def create_hermes_phot_table_row(datum, **kwargs):
         phot_table_row['brightness'] = datum.value['magnitude']
     else:
         phot_table_row['limiting_brightness'] = datum.value.get('limit', None)
-    if datum.value.get('magnitude_error', None):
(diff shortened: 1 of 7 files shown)


jchate6 (approved):
Thanks for this. I know it's been on the list for a while.

PR #913, opened 2024-04-29
add magnitude_error compatability

jchate6:
This adds some backwards compatibility for functions that previously relied on `magnitude_error` that have been now changed to use `error` instead.


phycodurus (approved):
I think comments in `hermes.py::L#165` and `dataproduct_extras.py::L#179` explaining why we handle both would be nice. (but whatever you decide, just merge it (i.e. consider it approved either way :-)

PR #921, opened 2024-05-09
Updated format of old-style links to other pages of documentation

rachel3834:
A user reported to me by private email that the links in the FAQ to other pages in our own docs were broken.  Looks like this is due to them not having been updated to the current URL formatting, so this is now done.


jchate6 (approved):
We really need to take some time and review all of the docs....

PR #922, opened 2024-05-09
clarify forced photometry vs catalog search

jchate6:
This generalizes the forced photometry infrastructure so that it can be used for any single target data service.


jnation3406 (approved):
Looks fine, mostly renaming stuff. Just out of curiosity, are you planning to have other non single target data services? Because otherwise I would've just called everything DATA_SERVICES instead of SINGLE_TARGET_DATA_SERVICES since it is less text everywhere and means the same thing...

jchate6:
@jnation3406 yes, we plan to have multi-target data services as well, similar to how Brokers work now, but more generalized.

phycodurus (approved):

PR #924, opened 2024-05-14
Feature/django htmx

phycodurus:
This PR
1. adds `htmx.min.js` in `tom_common/static/tom_common/js`
2. installs `htmx` in to the project via the <script> element in `base.html`

This makes htmx available for use anywhere site-wide (in particular, to the
tom_eso Facilty module.

`htmx.min.js` is loaded from a file (vs. from a CDN) for security in production.
(See comment in `base.html`)


tom_common/templates/tom_common/base.html
```diff
@@ -15,6 +15,10 @@
     <link rel="icon" type="image/x-icon" href="{% static 'tom_common/img/favicon-32.ico' %}" sizes="32x32" />
 
     {% bootstrap_javascript jquery='True' %}
+    <!-- there are security risks loading from CDNs. Do not do this in production -->
```
jchate6, line 18: Should this be concerning? Is there a solution for when a TOMToolkit is in production?

jchate6 (commented):
Can you add some links or something to `customization/customize_templates` in the docs to help people get started incorporating htmx into their custom templates?

tom_common/templates/tom_common/base.html
phycodurus: You can load from the CDN in development, but that is commented out on line 19. Line 20 explains that the htmx library is part of the project and it gets loaded from there on line 21.  So, maybe lines 18 and 19 should just be deleted as unnecessary bread crumbs.

jchate6 (approved):

PR #941, opened 2024-05-22
Fetaure/installed app api

### LINKED ISSUE #745: Add integration points to templates so additional modules can be added without overriding templates

phycodurus:
DRAFT: this refactor creates the infrastructure for generalized integration points for tom_base to introspect a TOMToolkit INSTALLED_APP. It's worth looking at these commits in isolation because they just set things up without adding any additional, complicating functionality. (That comes next).


jchate6:
@phycodurus I think we have taken this a different direction. Would you be OK Closing this PR?

tom_setup/templates/tom_setup/settings.tmpl
jchate6, line 1: We shouldn't need to include these specifically. They should run automatically when the app is installed.

PR #956, opened 2024-06-10
Add hermes spectro data sharing

jnation3406:
This adds support for spectra ReducedDatum hermes sharing in two formats, one of which is the format snex uses to store their spectra data.

Also renamed some html elements on the photometry datalist since i have to have similar code in snex2 for a spectroscopy datalist and they are all loaded into a single page, so better to be more specific with the javascript function names and stuff.


tom_dataproducts/alertstreams/hermes.py
```diff
@@ -168,6 +181,59 @@ def create_hermes_phot_table_row(datum, **kwargs):
...
+        raise HermesMessageException(msg)
+
+    spectroscopy_table_row = {
+        'target_name': datum.target.name,
+        'date_obs': datum.timestamp.isoformat(),
+        'telescope': datum.value.get('telescope', settings.DATA_SHARING['hermes'].get('DEFAULT_TELESCOPE', '')),
+        'instrument': datum.value.get('instrument', settings.DATA_SHARING['hermes'].get('DEFAULT_INSTRUMENT', '')),
+        'flux': flux_list,
+        'wavelength': wavelength_list,
+        'flux_units': datum.value.get('flux_units', settings.DATA_SHARING['hermes'].get('DEFAULT_FLUX_UNITS', 'mJy')),
```
jchate6, line 227: We should do a check here for classic [astropy units](https://docs.astropy.org/en/stable/units/ref_api.html#module-astropy.units.si) that have been converted to strings and convert them into what Hermes is looking for. (it probably makes more sense to accept astropy units in hermes and do any conversions there, just like we do for dates, but that might be outside the scope of this PR.)

tom_dataproducts/alertstreams/hermes.py
```diff
@@ -77,20 +84,24 @@ def preload_to_hermes(message_info, reduced_datums, targets):
         return response.json()['key']
     except Exception as ex:
         logger.error(repr(ex))
-        logger.error(response.json())
+        logger.error(response.content)
```
jchate6, line 87: We should add a similar log message (at least) up on line 69.

tom_dataproducts/sharing.py
```diff
@@ -280,13 +279,14 @@ def sharing_feedback_handler(response, request):
...
+        response.raise_for_status()
         if 'message' in response.json():
             publish_feedback = response.json()['message']
         else:
-            publish_feedback = f"ERROR: {response.text}"
+            publish_feedback = "Submitted message succesfully"
     except AttributeError:
         publish_feedback = response['message']
-    except ValueError:
+    except Exception:
```
jchate6, line 289: We could probably have this give some additional feedback based on the response content.

jchate6 (changes requested):
Made a few comments. Happy to talk about them in more detail if necessary.
Another thing to consider: There aren't currently many ways to access this spectroscopy sharing for Hermes. We should probably add spectra data to the main sharing page `target_share.html` and the spectroscopy tab of the `target_detail` page.

docs/managing_data/stream_pub_sub.rst
```diff
@@ -28,11 +28,15 @@ for the various streams with which you wish to share data.
...
-           'DISPLAY_NAME': os.getenv('HERMES_DISPLAY_NAME', 'Hermes'),
-           'BASE_URL': os.getenv('HERMES_BASE_URL', 'https://hermes.lco.global/'),
-           'HERMES_API_KEY': os.getenv('HERMES_API_KEY', 'set HERMES_API_KEY value in environment'),
-           'DEFAULT_AUTHORS': os.getenv('HERMES_DEFAULT_AUTHORS', 'set your default authors here'),
-           'USER_TOPICS': ['hermes.test', 'tomtoolkit.test']  # You must have write permissions on these topics
+            'DISPLAY_NAME': os.getenv('HERMES_DISPLAY_NAME', 'Hermes'),
+            'BASE_URL': os.getenv('HERMES_BASE_URL', 'https://hermes.lco.global/'),
+            'HERMES_API_KEY': os.getenv('HERMES_API_KEY', 'set HERMES_API_KEY value in environment'),
+            'DEFAULT_AUTHORS': os.getenv('HERMES_DEFAULT_AUTHORS', 'set your default authors here'),
+            'DEFAULT_TELESCOPE': '1meter',  # default telescope to fill in for datums missing telescope key
```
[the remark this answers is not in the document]
jnation3406: Yeah this was just for certain TOMs (and testing) that get their data predominantly from one source but don't put it in there reduced datums. If its left blank (not specified), then it will be empty and rejected from Hermes if its a required field (like telescope or instrument are).

docs/managing_data/stream_pub_sub.rst
```diff
@@ -28,11 +28,15 @@ for the various streams with which you wish to share data.
...
+            'DEFAULT_INSTRUMENT': 'inst01',  # default instrument to fill in for datums missing instrument key
+            'DEFAULT_WAVELENGTH_UNITS': 'Å',  # default units of wavelength for spectral data in this TOM
+            'DEFAULT_FLUX_UNITS': 'mJy',  # default units of flux for spectral data in this TOM
```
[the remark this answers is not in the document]
jnation3406: Yeah hmmm is the issue that I am suggesting values in the docs here? I assumed users would input their own values in their own TOMs, or have it empty if they don't want a default? Maybe I should explicitly leave those values empty in the docs then?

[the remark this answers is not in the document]
jchate6: I think my main issue is that this should be done somewhere else, like the dataprocessor. 
It feels like if we were going to apply a default value for any of these things it should be done at the `ReducedDatum` creation rather than at the point you are trying to share it. It feels like the "right" way to handle this is to force a user to consciously decide to apply a default at the moment of sharing rather than just share potentially incorrect units.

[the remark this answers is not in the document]
jnation3406: Yeah that sounds reasonable... I guess I need to add an extra step in submission where it detects some of these commonly missing pieces of information and prompts the user to fill in a field before sharing, since just failing saying there is no instrument or telescope isn't super helpful either (and there isn't an easy way to add that information into the TOM after the fact from the UI. I said in my talk I wanted to improve TOM integration for future work so this will be part of that :)

tom_dataproducts/alertstreams/hermes.py
jnation3406: So I added conversion functions for the units that existed in astropy, but none of them overlapped for flux... Unless I'm missing a way in the astropy units to denote like milli-, giga-, terra- to a unit. Also the default flux unit you use in the dataprocessor in the TOM is one we support in hermes but doesn't appear to be a unit in astropy unless I am missing it - it is an amalgamation of units.

docs/managing_data/stream_pub_sub.rst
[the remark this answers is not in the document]
jnation3406: I removed all the setting of default values, so it will be up to users to modify their data to include the required things to share. In the future I would like to prompt users to enter missing fields on submission, but that is much more involved then what I can do now.

jnation3406:
I've addressed most of your comments. It appears to me we just show a single plot of all the spectras on it in the spectra page. Should I add a button to share them all then, since we don't have any breakdown of them individually like in snex2? Honestly tom_base should probably pull in the snex2 spectra page to show the combined and all the individual ones with metadata since that seems more useful.

tom_dataproducts/alertstreams/hermes.py
jchate6: yeah, so a combined unit is still recognized as a unit. For example:
```
q = 42 * u.erg / u.s / u.cm / u.cm / u.Angstrom
q.unit
```
resolves to 
`Unit("erg / (Angstrom s cm2)")`

and `mJy` is a valid and understood astropy unit.

jchate6:
> I've addressed most of your comments. It appears to me we just show a single plot of all the spectras on it in the spectra page. Should I add a button to share them all then, since we don't have any breakdown of them individually like in snex2? Honestly tom_base should probably pull in the snex2 spectra page to show the combined and all the individual ones with metadata since that seems more useful.

I think implementing a SNex2 like spectra page is definitely the way to go. 
I imagine that people who want to share a spectrum with Hermes are usually only interested in sharing 1 at a time.
Andy might have a different opinion. 
For now, it might make more sense to have people share the data products related to spectra directly rather than from the spectrosopy page until we get a table with individual spectra listed?

jchate6 (approved):

jnation3406:
I've added a fix for the data sharing download functionality to work with spectrograph type reduced datums

tom_dataproducts/sharing.py
```diff
@@ -310,12 +342,20 @@ def download_data(form_data, selected_data):
     """
     reduced_datums = ReducedDatum.objects.filter(pk__in=selected_data)
     serialized_data = [ReducedDatumSerializer(rd).data for rd in reduced_datums]
+    data_to_save = []
```
jchate6, line 345: Can you update the doc string to reflect the spectroscopy changes?

jchate6 (changes requested):

jchate6 (approved):

PR #961, opened 2024-06-18
exclude identical object in ReducedDatum.validate_unique

griffin-h:
Right now, it's not possible to update an existing `ReducedDatum`, because its `validate_unique` method will always find *itself* as a duplicate.


jchate6 (approved):
good catch!

Issue #248, opened 2020-02-11
Add functionnality to download ReduceDatum not ingested with files


ebachelet:
I would like to have a button on the photometry target display that allow me to download the photometry, let say in json format for example. If it is ingested automatically in the db with ReduceDatum, i.e no file, then it is currently impossible to download the data from the website.

PR #962, opened 2024-06-18
download ReducedDatum not ingested with files

griffin-h:
This takes advantage of the data sharing form to provide a method of downloading photometry to a text file. Right now there's no way to download data not ingested via a file, as noted in #248. This may not be the preferred way to implement this feature request, but I'm opening a PR in case any part of the code I wrote is useful, either as an interim solution or to be reused in a better solution. Also feel free to close without merging!

To use this feature, select data from the sharing form (or select all), choose "download" as the destination, and click "submit." The "title" field will become the filename, and the "message" field will become a comment at the top of the file. The output table contains all the fields in the `ReducedDatum` table, not all of which are useful, including the superset of all keys in the `value` `JSONField`.


tom_dataproducts/views.py
```diff
@@ -30,7 +30,7 @@
 from tom_dataproducts.data_processor import run_data_processor
 from tom_observations.models import ObservationRecord
 from tom_observations.facility import get_service_class
-from tom_dataproducts.sharing import share_data_with_hermes, share_data_with_tom, sharing_feedback_handler
+from tom_dataproducts.sharing import share_data_with_hermes, share_data_with_tom, sharing_feedback_handler, download_data
```
jchate6, line 33: Can you split this line to make Lint Happy?

tom_dataproducts/sharing.py
```diff
@@ -293,3 +298,21 @@ def sharing_feedback_handler(response, request):
     else:
         messages.success(request, publish_feedback)
     return
+
+
+def download_data(form_data, selected_data):
```
jchate6, line 303: Can we get a short docstring describing this function?

jchate6 (changes requested):
This is great. Thanks for this.
Just a few very minor changes

tom_dataproducts/views.py
griffin-h: Done!

tom_dataproducts/sharing.py
griffin-h: Done! Let me know if it's clear.

jchate6: excellent. Thanks.

jchate6 (approved):
very nice.

PR #976, opened 2024-07-03
975 targetlists broken with custom user model

jchate6:
This is a fix for the TargetGroupings page when a custom Target Model is active.


rachel3834 (approved):
I've nothing to add here, this seems to make sense.

phycodurus (approved):

PR #1004, opened 2024-08-05
742a htmx for name select choice field

phycodurus:
This PR removes the Submit button and make changing the target name_select ChoiceField update the target fields table:

1. HTMX attributes were added to the name_select ChoiceField
2. a GET method was added to the TargetMergeView - if it's an HTMX request triggered by a name_select ChoiceField change, then it updates the table by calling the inclusiontag directly to get it's context and rendering the partial with the new context and return that new HTML. The htmx attributes of the ChoiceField specify (via that hx-target attribute) that the #id_target_merge_fields HTML should be replace.
3. the #id_target_merge_fields property was assigned to the <table> in the target_fields.html partial.


sglasberg (changes requested):
When I tried to run this I got an attribute error: 
`AttributeError: 'WSGIRequest' object has no attribute 'htmx'`

here is the traceback:
```
File "/home/sglasberg/dev/TOM_explore/env/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/sglasberg/dev/TOM_explore/env/lib/python3.11/site-packages/django/core/handlers/base.py", line 199, in _get_response
    response = self.process_exception_by_middleware(e, request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sglasberg/dev/TOM_explore/env/lib/python3.11/site-packages/django/core/handlers/base.py", line 362, in process_exception_by_middleware
    response = middleware_method(request, exception)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sglasberg/dev/tom_base/tom_common/middleware.py", line 29, in process_exception
    raise exception
  File "/home/sglasberg/dev/TOM_explore/env/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sglasberg/dev/TOM_explore/env/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sglasberg/dev/TOM_explore/env/lib/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sglasberg/dev/tom_base/tom_targets/views.py", line 641, in get
    if request.htmx:
       ^^^^^^^^^^^^
AttributeError: 'WSGIRequest' object has no attribute 'htmx'
```

sglasberg (changes requested):
please add `django_htmx` to `pyproject.toml` so that we install this dependency 

Could you also please update `settings.tmpl` with the required intalled app and middleware

tom_targets/views.py
```diff
@@ -603,30 +608,65 @@ def get_context_data(self, *args, **kwargs):
...
+        If this is not an HTMX request, just call super().get.
+        """
+        if request.htmx:
+            pk1 = int(self.kwargs.get('pk1', None))
+            pk2 = int(self.kwargs.get('pk2', None))
+
+            # get the target_id of the selected target: it's the primary
+            primary_target_id = request.GET.get('name_select', None)
+
+            # decide which of pk1 or pk2 is primary (i.e. it matches name_select)
```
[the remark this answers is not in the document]
phycodurus: Same :-< (I see that now, but I'm pretty sure that was working (locally) before -- looking into it...)

tom_targets/templatetags/targets_extras.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+import logging
```
sglasberg, line 2: Logging is not currently used in `targets_extras` should we remove this until it is needed

sglasberg (changes requested):
Looks pretty good

tom_targets/templatetags/targets_extras.py
phycodurus: Please leave this in.

sglasberg (approved):

PR #1008, opened 2024-08-08
742 manually link two targets as the same target and merge namesdatasets

### LINKED ISSUE #742: Manually Link Two Targets as the same target and merge names/datasets
When two different targets are discovered to be the same target, there should be a way to merge them and their data.

I feel like this should actually create a new target (in a target creation page) with the information from both targets filled in. 
Merging Target information (like RA/DEC) could be tricky, but the names would just become aliases, and all of the Reduced Datums and Data Products would be combined.

### Requirements:

- [x] User Interface
  -  Should be accessed from the Target List Page
  -  Only available when two targets are selected
- [ ] Management Command
  - Needs to take Target information in a variety of ways (name match, coord search, PK) and pare down to only 2.
- [ ] Back-end Code
  - [x] Quick option -- Takes "Preferred" target and uses that Target's values for any conflicting values, merging values where possible
  -  [ ] slow option -- each conflict must be resolved individually.
  - mergable parameters by default include: aliases, target lists, permissions, reduced datums, data Products
  - The less preferred name should be made into an alias
  - Handle and test for permission issues

sglasberg:
I finished my project, but may add more depending on time.


tom_targets/templates/tom_targets/partials/target_fields.html
```diff
@@ -0,0 +1,35 @@
...
+                        <td>{{ value1 }}</td>
+                    {% else %}
+                        <td>{{ value2 }}</td>
+                    {% endif %}
+                </tr>
+                {% endif %}
+            {% endif %}
+        {% endfor %}
+    </tbody>
+</table>
```
jchate6, line 35: Need a blank line and the end of the file

tom_targets/templatetags/targets_extras.py
```diff
@@ -285,6 +289,32 @@ def aladin_skymap(targets):
     return context
 
 
+@register.inclusion_tag('tom_targets/partials/target_fields.html')
+def target_fields(target1, target2):
+    """
+    Prepares the context for the target_fields.html partial.
+    Make a list of tuples that combines the fields and values of the two targets.
```
jchate6, line 296: describe the input for this method, Specifically what is expected from `target1` and `target2`

tom_targets/tests/tests.py
```diff
@@ -1727,3 +1731,192 @@ def test_share_empty_group(self):
...
+        self.user = User.objects.create(username='testuser')
+        self.st1 = SiderealTargetFactory.create()
+        self.st2 = SiderealTargetFactory.create()
+
+    def test_merge_targets(self):
+        self.st1.parallax = 3453
+        self.st1.save()
+        self.st2.distance = 12
+        self.st2.save()
+        # print(model_to_dict(self.st1))
```
jchate6, line 1747: Can remove this

tom_targets/tests/tests.py
```diff
@@ -1727,3 +1731,192 @@ def test_share_empty_group(self):
...
+        value2["magnitude"] = 12
+        ReducedDatum.objects.create(
+            target=self.st2,
+            data_type=data_type,
+            value=value2
+            )
+
+        st2_reduceddatums = list(ReducedDatum.objects.filter(target=self.st2))
+
+        #  write a test that makes sure datum1 and datum3 are the only datums associated with self.st1 after the merge
```
jchate6, line 1873: We decided not to do this yet...

tom_targets/views.py
```diff
@@ -561,6 +567,113 @@ def render_to_response(self, context, **response_kwargs):
...
 
+class TargetMergeView(FormView):
+    """
+    View that handles choosing the primary target in the process of merging targets
+    """
+
+    template_name = 'tom_targets/target_merge.html'
+    form_class = TargetMergeForm
+
+    def get_name_select_choices(self, pk1, pk2):
```
jchate6, line 578: Short doc string about what this method does and what it's inputs/outputs should be.

tom_targets/views.py
```diff
@@ -561,6 +567,113 @@ def render_to_response(self, context, **response_kwargs):
...
+
+        if form.is_valid():
+            primary_target_id = int(form.cleaned_data['name_select'])
+            if primary_target_id == first_target_id:
+                secondary_target_id = second_target_id
+            else:
+                secondary_target_id = first_target_id
+            primary_target = Target.objects.get(id=primary_target_id)
+            secondary_target = Target.objects.get(id=secondary_target_id)
+            if 'confirm' in request.POST:
```
jchate6, line 631: comment here about what happens if confirm is in the request

jchate6 (changes requested):
Mostly documentation stuff.

tom_targets/templates/tom_targets/partials/target_fields.html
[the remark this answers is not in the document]
sglasberg: Renamed to `target_merge_fields.html` will this break any of the code?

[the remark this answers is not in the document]
jchate6: yes, you'll need to change the name in target_extras

tom_targets/templatetags/targets_extras.py
```diff
@@ -285,6 +289,32 @@ def aladin_skymap(targets):
     return context
 
 
+@register.inclusion_tag('tom_targets/partials/target_fields.html')
```
jchate6, line 292: This will need to be changed if target_fields is changed

tom_targets/templatetags/targets_extras.py
```diff
@@ -285,6 +289,32 @@ def aladin_skymap(targets):
...
+def target_fields(target1, target2):
```
jchate6, line 293: If you make this name more specific, you'll need to update line 12 in target_merge.html too.

jchate6:
@sglasberg Please put back in the lines where you edit and save st1 and st2 in `test_merge_targets` the only thing we wanted to remove was the print statement.

tom_targets/merge.py
```diff
@@ -0,0 +1,90 @@
...
+    :type target: tom_targets.models.Target
+
+    :param secondary_target: Target object which holds all the secondary_target attributes
+    :type target: tom_targets.models.Target
+
+    returns: primary_target
+    """
+
+    model_fields = primary_target._meta.fields
+    """
```
jchate6, line 31: Convert this from a doc string to 2 lines commented normally `#` instead of `""""`

tom_targets/views.py
```diff
@@ -33,21 +35,25 @@
...
 from tom_dataproducts.sharing import (share_data_with_hermes, share_data_with_tom, sharing_feedback_handler,
                                       share_target_list_with_hermes)
 from tom_dataproducts.models import ReducedDatum
 from tom_targets.groups import (
     add_all_to_grouping, add_selected_to_grouping, remove_all_from_grouping, remove_selected_from_grouping,
     move_all_to_grouping, move_selected_to_grouping
 )
+from tom_targets.merge import (merge_error_message)
 from tom_targets.models import Target, TargetList
+from tom_targets.templatetags.targets_extras import target_fields
```
jchate6, line 50: this needs to be updated with the new name as well.

jchate6 (changes requested):
a few more things!

tom_targets/views.py
```diff
@@ -561,6 +567,117 @@ def render_to_response(self, context, **response_kwargs):
...
+            else:  # second is primary, so
+                secondary_target_id = pk1
+
+            # get the actual Target instances for these target_ids
+            primary_target = Target.objects.get(id=primary_target_id)
+            secondary_target = Target.objects.get(id=secondary_target_id)
+
+            # render the table with those targets via the inclusiontag
+            target_table_html = render_to_string(
+                'tom_targets/partials/target_fields.html',
```
jchate6, line 671: Here too

tom_targets/views.py
```diff
@@ -561,6 +567,117 @@ def render_to_response(self, context, **response_kwargs):
...
+                context=target_fields(primary_target, secondary_target))
```
jchate6, line 672: and here

jchate6 (changes requested):

tom_targets/views.py
sglasberg: flake is saying its unused should I delete this line

jchate6 (approved):

tom_targets/views.py
```diff
@@ -561,6 +567,117 @@ def render_to_response(self, context, **response_kwargs):
...
+        initial = {'target1': first_target,
+                   'target2': second_target}
+
+        form = TargetMergeForm(initial=initial)
+        form.fields['name_select'].choices = self.get_name_select_choices(first_target_id, second_target_id)
+        context['form'] = form
+
+        return context
+
+    def get_form_class(self):
```
phycodurus, line 613: This method is actually not necessary and could be deleted.

The reason is because the `form_class` is defined as a class variable (up on line #576) and the [super-class implementation](https://ccbv.co.uk/projects/Django/5.0/django.views.generic.edit/FormMixin/#get_form_class) returns that. So, the `form_class` could be define once and only once (DRY - don't repeat yourself principle at work here) up on line #576.

phycodurus (changes requested):

phycodurus (approved):

PR #1012, opened 2024-08-12
add guiding to LCO imaging facility page

jchate6:
I think this will do what is needed. I've tested it and submissions to the portal seem correct.


jchate6:
@rachel3834 this should be accessible from 2.19.5a2

rachel3834 (approved):
Looks good, I think this should serve our needs.

Issue #752, opened 2023-11-18
Current Visibility tracks for 1m LCO telescopes do not account for hour angle limits


gterreran:
When looking at the visibility tracks for LCO telescopes, some targets appear to be observables from 1m locations earlier (or later) than they actually are. Although the tracks are correct, and the targets will be at that altitude at the time, they do not factor the hour angle limits proper of the 1m mounts (this applies also to 0.4m, but not to 2m since they have alt-azimuth mounts). The hour angle limit is -4.6 hours and +4.6 hours (from the 1m telescope [pages](https://lco.global/observatory/telescopes/1m0/)).

Therefore when calculating the visibility tracks, these limits should be included and the tracks not be displayed when the target is outside the hour angle limits.

jchate6:
@gterreran So, this is a little tricky.
We don't display visibilities by telescope, we display them by site. 
We have no method right now in the TOM Toolkit to restrict object visibility to a single telescope or instrument, but only based on the celestial positions of the target and the Sun.

You will run into a similar issue if you try to look at a target that is past a pole or at the zenith of an alt az. 

Including hour angle limits (along with other telescope specific visibility restrictions) would effectively require us to plot each telescope per site rather than each site once. I'm not sure this would improve plot clarity rather than just make it busier and harder to read. For now a better solution may be to make sure we provide a good feedback message if you get a "target not visible within given window" message from the observing portal. In the future as we move to more reactive features, having the plot update with limits when you select the given instrument would be a great enhancement, but until then, does a more explicit feedback message sound like a good compromise?

gterreran:
@jchate6 Perhaps adding the information in the thumbnail? But yes, the error that the portal returns right now (particular the one parsed by the TOM) is confusing, because the object is indeed visible, it's the telescope that cannot reach it. So a more precise error explanation would definitely be beneficial to the users.

jchate6:
Consider using already available LCO resources: https://github.com/LCOGT/visibility or OCS plots

jchate6:
See also 
- #741

jchate6:
Additional airmass discussion can be found in 
- #1275

PR #1018, opened 2024-08-19
update LCO facility coordinates

jchate6:
Add decimal precision uniformity for LCO site coordinates and fix accuracy typo.

These errors were likely not enough to cause any substantial change in observation timings. The likely culprit for those is HA limits described in #752


phycodurus (approved):
Ideally, OCS would provide API to get these values and serve as a single source of truth. In the meantime :+1:

PR #1021, opened 2024-08-20
Updated name of config for single target data services

rachel3834:
If you follow the documentation for [integrating a single target data service](https://tom-toolkit.readthedocs.io/en/stable/managing_data/single_target_data_service.html) with a TOM, the name of the corresponding configuration dictionary in settings.py seems to have been changed.  This PR makes the code consistent with the docs, which use the more generalized data services terminology.


jchate6 (approved):
good catch. sorry I missed this in the refactor

PR #1026, opened 2024-08-22
Revised PanSTARRS data service to match recent API updates

rachel3834:
The PanSTARRS catalog service through MAST recently updated its API.  

Most notably, the fields for start and end timestamps seem to have been removed, and replaced with pagination and formatting parameters. 

This update revises the TOM's side of the API to match.

### DIFF
--- a/tom_dataproducts/single_target_data_service/panstarrs_service/panstarrs.py
+++ b/tom_dataproducts/single_target_data_service/panstarrs_service/panstarrs.py
@@ -24,13 +24,9 @@ DEFAULT_CONE_SEARCH_RADIUS_DEGREES = 0.008333  # degrees
 
 class PanstarrsSingleTargetDataServiceQueryForm(stds.BaseSingleTargetDataServiceQueryForm):
 
-    min_date_mjd = forms.FloatField(
-        label='Min date (MJD):', required=False,
-        widget=forms.NumberInput(attrs={'class': 'ml-2'})
-    )
-    max_date_mjd = forms.FloatField(
-        label='Max date (MJD):', required=False,
-        widget=forms.NumberInput(attrs={'class': 'ml-2'})
+    pagesize = forms.IntegerField(
+        label='Page size:', required=True, initial=500,
+        help_text=('Maximum number of entries to return per page.')
     )
 
     min_detections = forms.IntegerField(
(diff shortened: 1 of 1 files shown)


jchate6 (commented):
@rachel3834 I just tested this yesterday and it worked fine.
I believe the pagination option is an alternative, not a replacement for the date search.
It's the equivalent of asking for "the top X results" instead of "results between Date X and Y". I think we can add this feature rather than replace the old one.
What target are you struggling with, and what are you expecting vs what data are you getting?

rachel3834:
I am doing a test with the example from their website (RA=14:03:12.583, Dec=+54:20:55.50) but the query data with a date range e.g.

```
request_data = {
        'ra': target.ra,
        'dec': target.dec,
         'radius': DEFAULT_CONE_SEARCH_RADIUS_DEGREES,
       'epochMean.min': min_date_mjd,
          'epochMean.max': max_date_mjd,
        'nDetections.min': query_parameters.get('min_detections'),
        }
```

Produced no output (the response.content is a zero-length string) whereas this parameter dictionary matches the one on their website for the new API and does produce results:
```
        request_data = {
            'ra': target.ra,
            'dec': target.dec,
            'radius': DEFAULT_CONE_SEARCH_RADIUS_DEGREES,
            'nDetections.gte': query_parameters.get('min_detections'),
            'pagesize': query_parameters.get('pagesize'),
            'format': 'csv'
        }
```

rachel3834:
This is what I'm seeing if I run the PanSTARRS catalog query with the original code

mop_panstarrs_query

But I've just run the same query via a terminal and you're right the old API still works.  So there is some other issue.  
I will keep investigating.

jchate6:
So I notice that search is just from the 2nd of August to today.
If I broaden the date range, I only find data from before 2015.
Do we know that there has been more recent data, and that it should be in the PS1-DR2 Mean Object Catalog?

image

jchate6:
I notice the DR2 was released in 2019. What's the most recent data we should expect to see in it?

rachel3834:
Yes, I was also testing with wider date ranges.  With more investigation I think I understand - I think the issue was in the data processor for MOP needing to handle AWS IO rather than the API itself.  Apologies for my confusion.  Since I've got this far, I guess we can just add the pagination in as an option.  I am backing out the changes and I'll just add the pagination.

PR #1032, opened 2024-08-23
Handle merge condition before checking target lists

jchate6:
We should refactor this view to be a general Target List form and differentiate the idea of "Target Groupings" or `TargetList` from the `TargetListView` and the "Target List Page"


phycodurus (approved):
Please add a comment that adds some bit of context for this change and perhaps also explains why the merge functionality is in a View called `TargetAddRemoveGroupingView`.

PR #1053, opened 2024-09-30
update docs with a brief description of AnonymousUser

jchate6:
This is just meant to acknowledge the `AnonymousUser` and explain its purpose in the permissions section of the docs.
Let me know if you think this is sufficient and in the best location.
image

### DIFF
--- a/docs/common/permissions.rst
+++ b/docs/common/permissions.rst
@@ -14,7 +14,7 @@ The TOM Toolkit provides a permissions system that can be used in two different 
 ---------------
 
 When you first establish your TOM, ``django-guardian`` will create an ``AnonymousUser`` as the default user for the
-TOM. ``AnonymousUser`` is a special user that is used to represent users who are not logged in and only have permission
+TOM. ``AnonymousUser`` is a special user that is used to represent users who are not logged in and only has permission
 to see targets that are associated with the ``public`` group by default. This user is important for establishing what
 permissions are available to users who are not logged in and should not be removed. You can modify the permissions of
 ``AnonymousUser`` by using the Django admin interface or the methods described below.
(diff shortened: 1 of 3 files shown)


rachel3834 (approved):
Clear explanation, I think it covers everything users need to know.

docs/common/permissions.rst
```diff
@@ -10,6 +10,17 @@ The TOM Toolkit provides a permissions system that can be used in two different
 ``TARGET_PERMISSIONS_ONLY`` boolean in ``settings.py``.
 
 
+`AnonymousUser`
+---------------
+
+When you first establish your TOM, ``django-guardian`` will create an ``AnonymousUser`` as the default user for the
```
phycodurus, line 16: Maybe this is the correct way to explain this in the context of Permissions, but I think it leaves out the role AnonymousUser plays in a more general Django context, independent of ``django-guardian``. That is described in the Django docs [here](https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#anonymoususer-object).

Basically, `AnonymousUser` is a class defined in `django.contrib.auth.models`. The `AuthenticationMiddleware` attaches a user property to http requests as they pass through. If no User is logged in, then an instance `AnonymousUser` is attached to the request.

phycodurus (changes requested):

docs/common/permissions.rst
jchate6: I think if we want to add a section about `AuthenticationMiddleware` that's valid, but we should find a better place to put it.

Does it make sense to have a general page about the middleware generally? Or why certain things are set by default in `settings.py`?

At some point this feels like just teaching people Django, and there are better docs for that.

phycodurus (approved):

PR #1063, opened 2024-10-01
767 new tom creation documentation

jchate6:
This adds a variety of minor documentation changes.

### DIFF
--- a/docs/introduction/manual_installation.rst
+++ b/docs/introduction/manual_installation.rst
@@ -11,15 +11,8 @@ Manually Installing the TOM Toolkit and Django
 
 If you prefer to build your system manually, here's how.
 
-Firstly, the TOM toolkit requires you to have Python >= 3.7 installed
+Firstly, the TOM toolkit requires you to have Python >= 3.8.1, < 3.12 installed
 on your machine.
-
-If you are using Python 3.6 and cannot upgrade to 3.7, install the
-``dataclasses`` backport:
-
-::
-
-   pip install dataclasses
 
 We recommend using a `virtual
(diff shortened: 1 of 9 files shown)


docs/introduction/manual_installation.rst
```diff
@@ -11,16 +11,9 @@ Manually Installing the TOM Toolkit and Django
...
 on your machine.
 
-If you are using Python 3.6 and cannot upgrade to 3.7, install the
-``dataclasses`` backport:
-
-::
-
-   pip install dataclasses
-
 We recommend using a `virtual
```
phycodurus, line 17: As long as we're editing these docs, can we be more emphatic about this recommendation? For example say, "strongly recommend", or "urge you to use...", or "you'd be silly not to use...". Something like that?

docs/introduction/manual_installation.rst
```diff
@@ -82,7 +75,8 @@ that it looks like this:
```
phycodurus, line 76: Would it make sense to explain what's happening here and to also mention not to do this on a project that is already a TOMToolkit-based TOM? For example, say something like: "The tom_setup app adds a single management command (tom_setup) to your Django project. When run (see below) the tom_setup management command basically transforms your Django project into a TOMToolkit project by removing itself and adding the base TOMToolkit apps (i.e tom_targets, tom_observations, tom_dataproducts, etc) to your INSTALLED_APPS list."

phycodurus (approved):

jchate6:
@phycodurus would you like me to make these changes, or would you like to push a commit for these?

phycodurus (approved):

PR #1068, opened 2024-10-08
Add customizable class for formatting targets and datums into hermes …

jnation3406:
…format

So the current problem is that sharing with HERMES requires certain fields for targets/photometry/spectroscopy. We have a base implementation to pull those fields out of what keys we think they might be under, but everyone's TOM is different and some people use different keys and other people (SNEX) use other related models.

So this PR adds a class that encapsulates the code that converts a Target or ReducedDatum model instance into the corresponding data dictionary in hermes format, so that this class can be extended by individual TOMs to match their data format. This is a bit more work for an individual TOM to configure if they need to, but it is flexible enough to support use cases like SNEX2 with their custom models for storing associated data.

### DIFF
--- a/docs/managing_data/stream_pub_sub.rst
+++ b/docs/managing_data/stream_pub_sub.rst
@@ -16,6 +16,12 @@ need to copy your Hermes API Key from your Hermes profile page. When hermes shar
 buttons to open your data in hermes with the form pre-filled - this is a good option if you want to make slight changes
 to your message or data before sharing.
 
+To customize what data is sent to hermes from your ReducedDatum or Target models, please re-implement your own
+``tom_dataproducts.alertstreams.hermes.HermesDataConverter`` and customize the `get_hermes_*` methods to pull out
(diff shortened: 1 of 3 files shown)


jchate6 (approved):
This looks great. 
I think there's some validation issue on the hermes end though?
It won't accept data with just a limiting magnitude. 
When I try to submit such data I get `'{"data":{"photometry":[{"brightness_unit":["This field may not be null."]}...` as a response. Opening the data in Hermes results in no such error. Let me know if you'd like me to make an issue.

jnation3406:
> This looks great. I think there's some validation issue on the hermes end though? It won't accept data with just a limiting magnitude. When I try to submit such data I get `'{"data":{"photometry":[{"brightness_unit":["This field may not be null."]}...` as a response. Opening the data in Hermes results in no such error. Let me know if you'd like me to make an issue.

Hermes will set a default value if none is supplied for `brightness_unit` and `limiting_brightness_unit`, so I'll modify the code to not send that field if its null.

PR #1085, opened 2024-10-22
Set user in facility before calling the get_form_classes_for_display.

bobeldering:
When the `get_form_classes_for_display` method is called (as introduced by #1071), the user is not set in the facility instance. This is done a few lines later. 

This patch moves the call to `facility.set_user` up to before the call to `get_form_classes_for_display`.


jchate6 (approved):
Looks good.
THanks!

PR #1088, opened 2024-10-25
1073 add copy button to target coordinates radec

jchate6:
image
image
image

This is a multi-purpose copy button that could be used anywhere in the TOM Toolkit.

The pop-up toast closes after 3 seconds, and could also be used anywhere.


.gitignore
```diff
@@ -36,6 +36,8 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+*env/
```
phycodurus, line 39: whatever you decide about these is fine.
If they get left in, let's move them upto line 24

phycodurus (approved):

PR #1090, opened 2024-10-29
1055 python 312

jchate6:
This PR removes support for python 3.8 and adds support for python 3.12.

Many dependencies do not have python 3.13 support yet, so we will wait to implement that.

### DIFF
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.11'  # numpy is a problem with 3.12
+        python-version: '3.11'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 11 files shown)


.github/workflows/pypi-release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.11'  # numpy is a problem with 3.12
+        python-version: '3.11'
```
phycodurus, line 17: Just to clarify: we've removed the comment about 3.12, but not bumped the `python-version` (left it at 3.11). Is that what's intended?

phycodurus (approved):

.github/workflows/pypi-release.yml
jchate6: yes.
There is no longer a problem with 3.12 because we do not need to support 3.8, but there is also no reason to change to 3.12 while it's still in bug fixes.

PR #1123, opened 2024-11-22
Fix Basename Parsing in Router Registration Method

davner:
This PR addresses an issue in the register method of the custom `SharedAPIRootRouter` class, where the `basename` was not being correctly parsed due to an error in the logical expression’s precedence. This led to the `basename` being incorrectly set to `None`. This was discovered due to our tom crashing in cases where `viewsets` do not have a queryset to fall back on for default `basename` generation.

### With the fix
```console
registering new views: args: ('runprocessor', <class 'goats_tom.views.run_processor.RunProcessorViewSet'>), kwargs: {'basename': 'runprocessor'}

viewset = <class 'goats_tom.views.run_processor.RunProcessorViewSet'>, basename = 'runprocessor'
```

### Without the fix
```console

registering new views: args: ('runprocessor', <class 'goats_tom.views.run_processor.RunProcessorViewSet'>), kwargs: {'basename': 'runprocessor'}
viewset = <class 'goats_tom.views.run_processor.RunProcessorViewSet'>, basename = None

  router.register(r"runprocessor", views.RunProcessorViewSet, basename="runprocessor")

  File "/Users/dan.avner/git-clones/tom_base/tom_common/api_router.py", line 18, in register

    basename = self.shared_router.get_default_basename(viewset)

  File "/Users/dan.avner/miniconda3/envs/goats-env/lib/python3.10/site-packages/rest_framework/routers.py", line 170, in get_default_basename

    assert queryset is not None, '`basename` argument not specified, and could ' \

AssertionError: `basename` argument not specified, and could not automatically determine the name from the viewset, as it does not have a `.queryset` attribute.
```


jchate6 (approved):
Good Catch!
Thanks!

PR #1126, opened 2024-11-23
1104 add useraffiliation to the base user model

jchate6:
This is the basis for code to add profile information to a User for tom_base.

### Current issues:

- [x] If Users already exist without an associated profile, the profile page will break.
- [ ] The Profile model should be able to be overwritten by a User
- [x] Profile Fields are not properly displayed on the profile page via "verbose name"

Solutions to nonexistant profiles:
- Catch `RelatedObjectDoesNotExist: User has no profile.` error and create profile.
  - If we want to use the signals, we can do this in `save_profile`. This has the benefit of causing every new user to automatically have a profile, and updating all old ones automatically as they log in. Downside is using signals which adds new code.
  - Otherwise we create the new profile when accessing the profile page in `user_extras`. Has the benefit of no "new" code, but feels like the wrong place for this type of functionality.
- Allow users to not have profiles:
  - Build all code under the assumption that some users do not have profiles. Seems hard and surprising for users who want to access profile information for themselves in custom code.
- Use signals and add management command that needs to be run to create profiles for existing users when updating to new version. Seems like an unnecessary burden on users.


tom_common/tests.py
```diff
@@ -38,6 +38,8 @@ def test_user_list(self):
 
     def test_user_create(self):
         user_data = {
+            'profile-TOTAL_FORMS': '1',
```
jchate6, line 41: We changed the user creation form to a formset, so we need to include the profile ManagementForm data here or it will get angry.

phycodurus (approved):

PR #1139, opened 2024-12-10
Add ability to ingest Solar System targets via the MPC Explorer interface

talister:
The Minor Planet Center's [MPC Explorer](https://data.minorplanetcenter.net/explorer/) API interface (used elsewhere in LCO for resolving designations I believe) has recently added support for downloading orbital elements  (but sadly not for comets yet... sad snowball...) as well as designations and observations.  This PR adds a new `Harvester` which supports querying the MPC Explorer to produce `Target`s. 
This interface doesn't import the _H,G_ (or the (unpopulated) newer _H,G1, G2_ or cometary _M1, M2, K1, K2_) magnitude values but neither does the existing `MPCHarvester` and there's nowhere to put in the base `Target` anyway; possibly this is meant to go into `TargetExtra` custom fields?

Anyhoo, over to the developers for initial review...


jchate6:
If you would like this to show up in the list of Harvesters, you should add it to the `DEFAULT_HARVESTER_CLASSES` list in `tom_catalogs.apps.py` .

jchate6 (changes requested):
Thanks for this! Let me know if you would like me to implement these changes!

talister:
> If you would like this to show up in the list of Harvesters by default, you should add it to the `DEFAULT_HARVESTER_CLASSES` list in `tom_catalogs.apps.py` .

I figured that since 1) the MPC Explorer end point doesn't contain comets yet (despite only returning elements in comet form...) and 2) the majority of TOM users would not be using this endpoint, it didn't seem to make a sensible default for the whole project (and I've verified that adding it to FOMO's `settings.py` **does** work

tom_catalogs/tests/harvesters/test_mpc.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @patch('requests.get')
+    def test_query_failure_no_object(self, mock_get):
+        """test query of non-existant object"""
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.content = [None,  200]
+        mock_get.return_value = mock_response
+
+        result = self.broker.query('123456P')
```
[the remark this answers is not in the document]
talister: I had trouble with the Mock and the `canary` actual query making it fail in a useful way and was unsure what the general policy was on incorporating and simulating tests on unreachable or 500 error-returning sites. If you could take a look at potential "predictably failable" queries @jchate6 or point me at other examples (the `Simbad` Harvester mocks a `InconsistentTableError` which isn't applicable), I can look at implementing something in addition as I agree that it would be good

tom_catalogs/tests/harvesters/data/test_65803_mpc_orb.json
```diff
@@ -0,0 +1,289 @@
...
+        "mpcorb_version": "0.4"
+    },
+    "system_data": {
+        "EclipticObliquityArcseconds": "84381.448",
+        "eph": "DE431",
+        "force_model": "AST17",
+        "refframe": "ICRF",
+        "refsys": "Ecliptic"
+    }
+}
```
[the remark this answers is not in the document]
talister: Machines pacified in [[ea900c4](https://github.com/talister/tom_base/commit/ea900c4f87c3a71b10be675de8acfbbf607415de)

tom_catalogs/harvesters/mpc.py
```diff
@@ -30,3 +34,72 @@ def to_target(self):
...
+                # Store everything other than the status code for now and for later parsing.
+                self.catalog_data = response.json()[0:-1]
+
+    def to_target(self):
+        target = super().to_target()
+        result = self.catalog_data[0]['mpc_orb']
+
+        target.type = 'NON_SIDEREAL'
+        target.scheme = 'MPC_COMET'
+        target.name = result['designation_data']['iau_designation']
```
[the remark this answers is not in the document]
talister: This should be addressed now in 2a4b73fa

tom_catalogs/harvesters/mpc.py
```diff
@@ -30,3 +35,83 @@ def to_target(self):
...
+                mean_anomaly = td * target.mean_daily_motion
+                # Normalize into 0...360 range
+                mean_anomaly = mean_anomaly % 360.0
+                if mean_anomaly < 0.0:
+                    mean_anomaly += 360.0
+                target.mean_anomaly = mean_anomaly
+
+        # Save Target, adding in additional names
+        # target.save(names=extra_desigs)
+        target.save()
```
jchate6, line 112: This bypasses the form.
Basically this submits the target during the search phase and populates the form with already saved information.
I understand why this is done like this, but what we actually want is to populate the alias fields in the form and save them after the form is submitted.

I'll work on figuring out how to pipe this through correcty.

jchate6 (changes requested):
Can you please add instructions on how to include the `MPCExplorerHarvester` via `settings.py` in the harvester doc string, and add the Harvester to the list (similar to JPL Horizons harvester) in `docs.api.tom_catelog.harvesters.rst`?

I'll take care of the form stuff. That might be more complicated.

tom_catalogs/harvesters/mpc.py
talister: This was the way I could get it work and make the tests pass without getting duplication (in the case of provisional id only) or missing ids. I'm sure it's not the only way though...

jchate6: This was much easier than I thought.

talister: It breaks the tests for me:
```
======================================================================
ERROR: test_to_target (tom_catalogs.tests.harvesters.test_mpc.TestMPCExplorerHarvester.test_to_target)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/eng/git/tom_base/tom_catalogs/tests/harvesters/test_mpc.py", line 34, in test_to_target
    self.assertEqual(target.names, ['65803', 'Didymos', '1996 GT'])
                     ^^^^^^^^^^^^
  File "/home/eng/git/tom_base/tom_targets/base_models.py", line 491, in names
    return [self.name] + [alias.name for alias in self.aliases.all()]
                                                  ^^^^^^^^^^^^^^^^^^
  File "/home/eng/venv/tombase311_venv/lib64/python3.11/site-packages/django/db/models/manager.py", line 164, in all
    return self.get_queryset()
           ^^^^^^^^^^^^^^^^^^^
  File "/home/eng/venv/tombase311_venv/lib64/python3.11/site-packages/django/db/models/fields/related_descriptors.py", line 718, in get_queryset
    raise ValueError(
ValueError: 'BaseTarget' instance needs to have a primary key value before this relationship can be used.
```

jchate6: yes. working on that.

PR #1140, opened 2024-12-10
fix profile bug

jchate6:
Somehow this code was accidentally removed before the release.

### DIFF
--- a/tom_common/apps.py
+++ b/tom_common/apps.py
@@ -8,6 +8,7 @@ class TomCommonConfig(AppConfig):
 
     def ready(self):
         # Import signals for automatically saving profiles when updating User objects
+        # https://docs.djangoproject.com/en/5.1/topics/signals/#connecting-receiver-functions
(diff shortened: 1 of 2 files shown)


tom_common/apps.py
```diff
@@ -7,6 +7,9 @@ class TomCommonConfig(AppConfig):
     name = 'tom_common'
 
     def ready(self):
+        # Import signals for automatically saving profiles when updating User objects
```
phycodurus, line 10: Is there a reason for not importing this at the top of the module?

phycodurus (approved):

tom_common/apps.py
jchate6: I am following the [docs](https://docs.djangoproject.com/en/5.1/topics/signals/#connecting-receiver-functions):

```If you’re using the receiver() decorator, import the signals submodule inside ready(), this will implicitly connect signal handlers```

PR #1142, opened 2024-12-19
Remove target input from form if its fixed, add ability to share targ…

jnation3406:
…et and its existing data when creating a persistent share, remove download from the destinations list, and bold the service portion of destination string. I had to refactor where some methods were to get around circular imports.

Many of these come from comments from Andy yesterday when he saw the sharing page. Here is a screenshot of what it looks like now:
image

### DIFF
--- a/tom_targets/sharing.py
+++ b/tom_targets/sharing.py
@@ -23,12 +23,14 @@ def share_target_and_all_data(share_destination, target):
         filtered_reduced_datums = check_for_share_safe_datums(
             destination, ReducedDatum.objects.filter(target=target), topic=hermes_topic)
         sharing = getattr(settings, "DATA_SHARING", {})
+        tom_name = f"{getattr(settings, 'TOM_NAME', 'TOM Toolkit')}"
         message = BuildHermesMessage(title=f"Setting up continuous sharing for {target.name} from "
-                                     f"{getattr(settings, 'TOM_NAME', 'TOM Toolkit')}.",
+                                     f"{tom_name}.",
                                      authors=sharing.get('hermes', {}).get('DEFAULT_AUTHORS', None),
-                                     message=None,
+                                     submitter=tom_name,
+                                     message='',
                                      topic=hermes_topic
-                                    )
+                                     )
         response = publish_to_hermes(message, filtered_reduced_datums)
         response.raise_for_status()
     else:
@@ -55,10 +57,12 @@ def continuous_share_data(target, reduced_datums):
             filtered_reduced_datums = check_for_share_safe_datums(
                 destination, ReducedDatum.objects.filter(pk__in=reduced_datum_pks), topic=hermes_topic)
             sharing = getattr(settings, "DATA_SHARING", {})
+            tom_name = f"{getattr(settings, 'TOM_NAME', 'TOM Toolkit')}"
             message = BuildHermesMessage(title=f"Updated data for {target.name} from "
-                                         f"{getattr(settings, 'TOM_NAME', 'TOM Toolkit')}.",
+                                         f"{tom_name}.",
                                          authors=sharing.get('hermes', {}).get('DEFAULT_AUTHORS', None),
-                                         message=None,
+                                         submitter=tom_name,
+                                         message='',
                                          topic=hermes_topic
                                          )
             publish_to_hermes(message, filtered_reduced_datums)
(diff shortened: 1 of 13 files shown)


tom_targets/sharing.py
```diff
@@ -4,7 +4,66 @@
...
 
 from tom_targets.serializers import TargetSerializer
-from tom_dataproducts.sharing import get_destination_target
+from tom_targets.models import PersistentShare
+from tom_dataproducts.sharing import check_for_share_safe_datums, share_data_with_tom, get_destination_target
+from tom_dataproducts.models import ReducedDatum
+from tom_dataproducts.alertstreams.hermes import publish_to_hermes, BuildHermesMessage
+
+
+def share_target_and_all_data(share_destination, target):
```
jchate6, line 13: This is silently breaking for me.
I'm getting back `b'{"message_text":["This field may not be null."],"submitter":["This field may not be blank."]}'`
Obviously these things need to be autofilled or use the other form.
It would also be nice if these made the same user facing alerts that normal sharing does (for success as well as failure.)

tom_targets/templates/tom_targets/partials/create_persistent_share.html
```diff
@@ -3,12 +3,17 @@
 <form method="post" class="form" id='target-persistent-share-create-form'>
   {% csrf_token %}
   <div class="form-row" style="padding-inline:1rem">
-      <div class="col-sm-3">
+      <div class="col-sm-4">
           {% bootstrap_field form.destination %}
       </div>
-      <div class="col-sm-5">
+      {% if not target %}
```
jchate6, line 9: Like this change a lot.

tom_targets/sharing.py
```diff
@@ -4,7 +4,66 @@
...
+        destination = share_destination.split(':')[0]
+        filtered_reduced_datums = check_for_share_safe_datums(
+            destination, ReducedDatum.objects.filter(target=target), topic=hermes_topic)
+        sharing = getattr(settings, "DATA_SHARING", {})
+        message = BuildHermesMessage(title=f"Setting up continuous sharing for {target.name} from "
+                                     f"{getattr(settings, 'TOM_NAME', 'TOM Toolkit')}.",
+                                     authors=sharing.get('hermes', {}).get('DEFAULT_AUTHORS', None),
+                                     message=None,
+                                     topic=hermes_topic
+                                    )
```
jchate6, line 31: missing a space before the parentheses here to make flake happy

jchate6 (changes requested):

tom_targets/sharing.py
jnation3406: Thanks for checking that - I had thought submitter would be autoset by Hermes but it actually doesn't do that for outgoing messages, only incoming. So I've set the submitter to be the `TOM_NAME` which seems appropriate to me, and set the message text to be empty string instead of None.

jnation3406: Also, the error should me mapped through now so it should stop the creation of a persistent share and show the error on screen if you have a problem when sharing all existing data. If it succeeds, nothing is shown but the persistent share is created and will show up, so that is the signal that it succeeded in sharing existing data too.

jchate6: I'm still getting a silent failure for `HTTPError('400 Client Error: Bad Request for url: https://hermes-dev.lco.global/api/v0/submit_message/')
b'{"topic":["Hermes Dev can only submit to the hermes.test topic."]}'`

as well as missing telescope/instrument

jnation3406: Hmm I'll look into the error not making it through... What do you mean by missing telescope/instrument though? Is it stored in your datums value dict as 'telescope' and 'instrument'?

jchate6: yeah, not having the telescope/instrument in my data is just another way my hermes message can be rejected. The issue isn't that, it's just one of the two ways I tried to get errors back from hermes.

jnation3406: Ah okay I understand what were saying now

jnation3406: I've updated it again, and the errors are definitely coming through now! Sorry the error feedback handler stuff had me confused a bit.

jchate6 (approved):
Looks good. Thanks!

PR #1143, opened 2024-12-20
1105 add integration points for apps to include partials on the users profile page

jchate6:
Allow other apps to integrate profile information onto the profile page.

image


tom_common/templatetags/user_extras.py
```diff
@@ -42,3 +46,37 @@ def user_data(user):
...
+                try:
+                    clazz = import_string(profile['context'])
+                except ImportError:
+                    logger.warning(f'WARNING: Could not import context for {app.name} profile from '
+                                   f'{profile["context"]}.\n'
+                                   f'Are you sure you have the right path?')
+                    continue
+                new_context = clazz(user)
+                for item in new_context:
+                    context[item] = new_context[item]
```
phycodurus, line 77: This forces each app to know about every other app's context and avoid key naming collisions. How about each app having it's own dictionary within the overall context? So, context = {'tom_demoapp: {'context': 'path.to.template.html',
                              'partial': 'path.to.inclusiontag-decorated-function'}
               'tom_swift': {like above example},
               <general context data>
               }
The key for the subdictionary should be the `AppConfig.name`

phycodurus (changes requested):

phycodurus (approved):
yay!

PR #1146, opened 2024-12-21
move profile page from user detail view

jchate6:
Prevents people from accessing other people's profiles.

### DIFF
--- a/tom_common/views.py
+++ b/tom_common/views.py
@@ -82,22 +82,11 @@ class UserDeleteView(LoginRequiredMixin, DeleteView):
             return super().dispatch(*args, **kwargs)
 
 
-class UserDetailView(LoginRequiredMixin, DetailView):
+class UserProfileView(LoginRequiredMixin, TemplateView):
     """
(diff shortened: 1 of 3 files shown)


tom_common/views.py
```diff
@@ -82,22 +81,11 @@ def dispatch(self, *args, **kwargs):
             return super().dispatch(*args, **kwargs)
 
 
-class UserDetailView(LoginRequiredMixin, DetailView):
+class UserProfileView(LoginRequiredMixin, TemplateView):
     """
     View to handle creating a user profile page. Requires a login.
     """
```
phycodurus, line 87: Let's add to the doc string that

- the urlpattern is `user/profile` because
  - this is not a DetailView subclass, hence no pk in url; also security
  - it is a LoginRequiredMixin sublass and that implies that the user is known as part
    of the request and that the template depends on that

Basically a comment that explains how the pieces (the template, the mix-in, the urlpattern, etc are interacting to make this work...

phycodurus (approved):

PR #1149, opened 2025-01-06
remove 'reply' layer from TNS 2.0 responses

griffin-h:
As of December 31, 2024, responses from the TNS API no longer have a 'reply' key.


jchate6:
related to:
 - TOMToolkit/tom_tns#39

jchate6 (approved):
Works Great. Thanks for this.

PR #1152, opened 2025-01-09
Update pyproject.toml file to be more universal and compatible with poetry 2.0

phycodurus:
This is a first step to fix the PyPI submission issue


jchate6 (approved):
nice work

PR #1153, opened 2025-01-13
1151 improve display of undefined values from jpl horizons

jchate6:
This PR does 2 things.

1. We convert the 9.999E+99 result from Horizons into a `None`
2. We finally fix the display issues with the Target detail data to not just be 4 decimal places for everything.

image


tom_catalogs/harvesters/jplhorizons.py
```diff
@@ -28,6 +28,10 @@ def to_target(self):
         target = super().to_target()
         target.type = 'NON_SIDEREAL'
         target.scheme = 'MPC_MINOR_PLANET'
+        # Remove awkward infinities from the catalog data
+        for column in self.catalog_data.columns.values():
+            if column[0] == 9.999999999999998e+99:
```
phycodurus, line 33: I'm not sure how this magic number is represented, but that `==` comparison might be a little brittle. Would `>=` or some other comparison (i.e. greater than x, and less than y?) be more robust?

phycodurus (approved):

tom_catalogs/harvesters/jplhorizons.py
jchate6: I don't think so. 
That's the magic number that Horizons uses if a value is infinity or doesn't apply.
We could do something else, but the brittleness is dominated by the Horizons API, and that should be caught in a canary test.

PR #1173, opened 2025-02-05
Add index on BaseTarget.created to speed up load time of large DBs

### LINKED ISSUE #598: TargetList load time for non-superusers
For non-superusers, some pages like the TargetList taks a long time to load, perhaps b/c of permissions checking

This is related to django-guardian checking.
https://django-guardian.readthedocs.io/en/stable/userguide/performance.html

Fingel:
We currently sort on the created field of BaseTarget, which causes performance issues when there are a large number of BaseTargets.

In a database with 6 million rows, this reduced the time for the SQL SELECT from ~800ms to less than 1ms, significantly speeding up page load time.

Note that this does NOT solve #598 but it does help a little.

No index:
Screenshot From 2025-02-05 13-58-41

With index:
image


jchate6 (approved):
I'm pretty sure this is what I did to make the home page accessible.
But yeah, it doesn't totally solve the problem.

PR #1175, opened 2025-02-06
Permissions attribute for targets, filtering logic and migrations

Fingel:
Adds a permissions field on the Target that allows for filtering prior to doing row-level permissions for open/public targets. This significantly improves performance for TOMs with large datasets that are mostly open or public.

Included migrations should handle the transition transparently.

Todo:

- [x] Extend filtering logic beyond the TargetList view. 
- [x] Update target creation logic away from the Public group. 
- [x] Apply logic to non-target models?
- [x] Tests.
- [x] Clean up any remaining 'Public' group logic.
- [x] Update documentation.


tom_targets/base_models.py
```diff
@@ -306,6 +311,10 @@ class BaseTarget(models.Model):
         auto_now=True, verbose_name='Last Modified',
         help_text='The time which this target was changed in the TOM database.'
     )
+    permissions = models.CharField(
+        max_length=100, default=Permissions.PUBLIC, choices=Permissions.choices,
```
Fingel, line 315: Should we set the default to private? It's possible people have built pipelines already with the assumption that newly created targets will have row-level permissions enforced.

tom_targets/templates/tom_targets/target_form.html
```diff
@@ -35,3 +34,6 @@ <h3>Tags</h3>
 {% endbuttons %}
 </form>
 {% endblock %}
+{% block extra_javascript %}
+<script defer src="https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js"></script>
```
Fingel, line 38: Alpine.js is going to be useful for some of the other stories relating to making the UI more responsive. I'm not up to date on how we are handling static assets, but at the least this can be moved to `base.html`.

tom_targets/seed.py
```diff
@@ -117,11 +115,7 @@
 
 
 def seed_messier_targets():
-    public, _ = Group.objects.get_or_create(name='Public')
     target_list, _ = TargetList.objects.get_or_create(name='Messier Catalog')
     for target in messier_targets:
         t, _ = Target.objects.get_or_create(name=target[0], ra=target[1], dec=target[2], type=Target.SIDEREAL)
```
Fingel, line 120: change this to public

tom_targets/templates/tom_targets/partials/sidereal_target_form.html
```diff
@@ -0,0 +1,79 @@
...
+<div class="row mb-3" x-data="{permission: 'PUBLIC' }">
+    <div class="col-md-6">
+        <select x-model="permission" name="permissions" class="form-control" id="id_permissions">
+          <option value="OPEN">Open</option>
+          <option value="PUBLIC">Public</option>
+          <option value="PRIVATE">Private</option>
+        </select>
+    </div>
+    <div class="col-md-6">
+        <div x-show="permission == 'PRIVATE'" x-transition>
```
Fingel, line 74: Add description of permissions here.

tom_targets/templates/tom_targets/partials/sidereal_target_form.html
```diff
@@ -0,0 +1,79 @@
...
+            <span>Target will only be visibile to members of these groups:</span>
+            {% bootstrap_field form.groups show_label=False %}
```
Fingel, line 76: Check if there are any groups here, display message if not.

jchate6:
@Fingel Didn't have a chance to look closely, but there seems to be an issue with when the "Add a group" section is displayed:

image

Fingel:
> @Fingel Didn't have a chance to look closely, but there seems to be an issue with when the "Add a group" section is displayed:
> 
> image

Strangely cannot reproduce this, but try with the latest commit - should be a better way to do this check.

jchate6:
> Strangely cannot reproduce this, but try with the latest commit - should be a better way to do this check.

Sorry, this was in the Target update view. The Target create view did work correctly. This change fixed it. Thanks!

tom_targets/templates/tom_targets/partials/non_sidereal_target_form.html
```diff
@@ -0,0 +1,85 @@
...
+        {% bootstrap_field form.ephemeris_epoch %}
+    </div>
+    <div class="col-md-4">
+        {% bootstrap_field form.ephemeris_epoch_err %}
+    </div>
+    <div class="col-md-4">
+        {% bootstrap_field form.perihdist %}
+    </div>
+</div>
+<div class="form-row">
```
jchate6, line 81: We can remove parallax. This shouldn't be here. This value has no real meaning for non-sidereal targets. 
( I know this was on the previous form... it shouldn't have been there either. :smile: )

jchate6 (changes requested):
Just a few things. Looks great though. Very excited for these changes.

tom_targets/base_models.py
Fingel: It's a database default, so while we could put it in settings that would cause a migration in tom_targets any time it's changed.

phycodurus (approved):

jchate6 (approved):

PR #1176, opened 2025-02-11
address current limitations in astroquery table join, and update tests

jchate6:


tom_catalogs/tests/harvesters/test_simbad.py
```diff
@@ -10,24 +9,25 @@
 class TestSimbadHarvester(TestCase):
     def setUp(self):
         self.broker = SimbadHarvester()
-        table_data = {'ra': [10.68470800], 'dec': [41.26875000],
-                      'pmra': ['--'], 'pmdec': ['--'], 'main_id': ['M 31, 2C 56, DA 21'],
-                      'mesdistance.dist': [0.8200], 'mesdistance.unit': ['Mpc']}
+        table_data = {'main_id': ['M  31'], 'ra': [10.684708333333333],
+                      'dec': [41.268750000000004], 'pmra': ['--'], 'pmdec': ['--'],
```
phycodurus, line 13: if it's not too much trouble, can we reformat this dict and clean up the ra,dec significant digits?

phycodurus (approved):

tom_catalogs/tests/harvesters/test_simbad.py
jchate6: Those are actual outputs from Simbad for M31. How would you like it to be re-formatted? 
I don't want to change the digits in the RA/Dec because that's what simbad outputs and this is testing that output.

phycodurus (approved):
thank you

Issue #1141, opened 2024-12-14
Users should not be able to see Target Groups  on the Target List page that they don't have permission to access


jchate6:
We should limit the list in the dropdown to only the lists users have created or otherwise have access to.
![image](https://github.com/user-attachments/assets/a8c8bcac-759f-4fdc-9358-7363895848db)
![image](https://github.com/user-attachments/assets/90cf7d3d-f64a-44c9-bccb-3ff07d92e195)

I think giving users the ability to add `User Groups` to a `Target Group` in a way similar to Targets would be a convenient way to share target groups across users who aren't super users.

Issue #679, opened 2023-08-09
Consolidate usage of "Target Group" vs "TargetList"


jchate6:
As far as I can tell we use the terms "Target List" and "Target Group" more or less interchangeably with "Group" tending to be more user facing than "list". This inconsistency can be confusing and add unnecessary complexity for TOM users and especially TOM admin. 

I'd recommend a refactor in which we choose a single term for this idea and use consistent variable names throughout the toolkit and documentation.

(I'd also like to see a similar refactor for alias vs target_name").

rachel3834:
I agree, and I'd favor the term targetlist rather than group, as in my
experience this is the more common parlance among astros.

jchate6:
My only counter arguments in favor of "Group" over "List" are that 

1. Group is currently the more user facing term while target list is mostly in the code.
2. A traditional target "list" as most astronomers would use the term is possibly only a subset of the use cases for target "Groups" as they exist in TOM Toolkit. Referring to these as "target lists" might lead to a very limited interpretation of what these are. In Tom Toolkit, a group is much more akin to a high level "tag" than what I would consider a traditional target list with each target being a member of many groups.

rachel3834:
Those are good points, especially point 2.  I do worry that currently users
have shown a little confusion over what the target groups are for at the
moment, for essentially the opposite reason: they were looking to create a
target list and didn't think about groups as a way to do this.  But perhaps
that's something we can overcome in the documentation - likely different
people would favor different options.  We could do a poll through the TOM
Slack channel and see what they vote for.

jchate6:
The complicating factor for all of this is the existence of `UserGroups` which are also called "groups" and `targetgroupings` in the URLs

jchate6:
Eliminate any use of "group" or "list" by themselves in the code.
These should be replaces with "ObjectGroup" where "object" is the model being grouped:
`TargetGroup`
`UserGroup`
`ObservtionGroup`
etc.

PR #1181, opened 2025-02-14
Visual representation of target groups

Fingel:
Adds representation of the groups a target belongs to in various parts of the UI. Groups can be assigned a color upon creation.

On the target list, the groups appear as pills beneath the target name. Clicking on a group pill will filter the table by that target group.

On the target detail page, clicking a group pill focuses the "manage groups" tab.

image

image


tom_targets/models.py
```diff
@@ -178,8 +178,23 @@ class TargetList(models.Model):
...
+        GREY = 'secondary'
+        GREEN = 'success'
+        YELLOW = 'warning'
+        RED = 'danger'
+        CYAN = 'info'
+        LIGHT = 'light'
+        DARK = 'dark'
+
     name = models.CharField(max_length=200, help_text='The name of the target list.')
+    color = models.CharField(max_length=20, choices=Color.choices, default=Color.BLUE)
```
jchate6, line 197: It feels like if we are going to do this, we should just use a colorpicker widget and store a hexcode or something.

tom_targets/models.py
```diff
@@ -178,8 +178,23 @@ class TargetList(models.Model):
 
     :param modified: The time at which this target list was modified in the TOM database.
     :type modified: datetime
+
+    :param color: The color used to represent this list in UI elements.
+    :type color: str
     """
+
+    class Color(models.TextChoices):
+        BLUE = 'primary'
```
jchate6, line 187: This notation is problematic. Part of the benefit of using "primary", "secondary", "info", etc. is that users can change their CSS themes and their buttons will be consistent. For users who have customized their colors, Primary will not necessarily be "Blue".

tom_targets/templates/tom_targets/partials/target_feature.html
```diff
@@ -10,6 +10,13 @@ <h3>{{ target.name }}
     {% endwith %}
 {% endif %}
 </h3>
+<div class="mb-1">
+{% for list in target.targetlist_set.all %}
```
jchate6, line 14: I don't love these here.
The badges seem distracting and can really pile up without any real benefit I can see.
I'd rather not have anything between the main buttons and the name if at all possible for clarity.
We have a manage groups tab that can be used to actually find other targets in the list and remove the target from a list.

tom_targets/templates/tom_targets/partials/target_table.html
```diff
@@ -30,6 +31,15 @@
       </td>
       <td>
           <a href="{% url 'targets:detail' target.id %}" title="{{ target.name }}">{{ target.names|join:", " }}</a>
+          <div class="target-lists">
+            {% for l in target.targetlist_set.all %}
+            <a href="{% url 'targets:list' %}?targetlist__name={{ l.id }}">
+                <span class="badge badge-pill badge-{{ l.color }}">
```
jchate6, line 37: I feel like a pill badge might be too limiting for these. 
I really feel like the colors are not flexible enough, and the solid pills are too bold compared to the name and the rest of the table. I'd rather a more subtle tag with just a border and a name similar to a `btn-outline-*` look.

tom_targets/templates/tom_targets/targetlist_form.html
```diff
@@ -2,8 +2,19 @@
 {% load bootstrap4 %}
 {% block title %}New Target Group{% endblock %}
 {% block content %}
-<form method="post">{% csrf_token %}
-    {{ form.as_p }}
-    <input type="submit" value="Create">
+<h4>New Target List</h4>
```
jchate6, line 5: The targetList List view should be updated to show the color, and we should be able to edit the TargetList if needed.

jchate6 (changes requested):
I love the attempt to modernize our target table, but I think a fundamental issue with this PR is that users have a huge variety of ways that they use TargetLists. These aren't just general groupings of targets like "messier objects" or "Interesting" or "Dave's pet targets". People also use these to mark targets that they observed or planned to observe during an observing run, targets tied to a specific proposal, or targets tat have had "super-delux"modeling algorythm 1.0.2" run on them as opposed to v 1.0.1. 

Also, targetLists are a per-user organizational contrivance that SHOULD NOT be accessed by other users (see issue #1141 ) who are not super-users. Currently all of these pills on both the target list and target detail pages are visible to everyone who can view the target.

The bottom line, is that I don't know if a tag like this is going to add functionality for most users, and could even cause significant problems depending on how a user has chosen to use their TargetLists. (You could imagine long descriptive list names, or having a target on hundreds of lists.)

Furthermore, there are other things related to target lists we'd like to clean up.
See issue #679 and #1033 

I think we should talk about some more nuanced method involving user preferences, only displaying a few recent tags, or having a toggle in the user profile to deactivate them all together if we want to pursue this feature.

Fingel:
@jchate6,
The target list page has a bulk option "Add/Remove targets from grouping" and unless I'm missing something, it is impossible to know which groups a target does or does not belong to without entering the target detail page - making any such bulk operations impossible, or just blind. So as far as I'm concerned the target list page is broken. I'm sure we can come up with a million ways why pills are not optimal, in the meantime the page is still broken. But hey I get it, it's been this way for 7 years why change it now.

PR #1183, opened 2025-02-20
Infinite scroll on target list table using HTMX

Fingel:
Here is how infinite scroll might work using HTMX.

The django-template-partials dependency is on track to be included in Django, hopefully in time for the 5.2 release.
https://github.com/carltongibson/django-template-partials/issues/29

The `querystring` template tag added to `tom_common_extras` is taken directly from Django 5.1, so once `tom_base` is upraded to 5.1 the entire tag can simply be removed.

What's left isn't much! There were some changes required because of using a tag for the target table instead of a template include, but nothing too crazy.

The select some/select all checkbox remains broken in the same way it is before this change, so that can be addressed in another PR.

https://github.com/user-attachments/assets/63e49a75-e85f-4182-b6bb-637c003527aa

### DIFF
--- a/tom_targets/templates/tom_targets/target_list.html
+++ b/tom_targets/templates/tom_targets/target_list.html
@@ -23,8 +23,7 @@
       </div>
     </div>
     {% select_target_js %}
-    {% aladin_skymap object_list %}
-    {% bootstrap_pagination page_obj extra=request.GET.urlencode %}
+    {% aladin_skymap targets %}
     <label id="displaySelected"></label>
     <button id="optionSelectAll" type="button" class="btn btn-link" onClick="select_all({{ target_count }})"></button>
     <form id="grouping-form" action="{% url 'targets:add-remove-grouping' %}" method="POST">
@@ -47,8 +46,7 @@
         <button type="submit" class="btn btn-outline-danger ml-1" name="remove">Remove</button>
       </div>
     </form>
-    {% target_table object_list %}
-    {% bootstrap_pagination page_obj extra=request.GET.urlencode %}
+    {% target_table targets %}
   </div>
   {{ filter.fields }}
   <div class="col-md-2">
--- a/poetry.lock
+++ b/poetry.lock
@@ -972,6 +972,25 @@ asgiref = ">=3.6"
 django = ">=4.2"
 
 [[package]]
+name = "django-template-partials"
+version = "24.4"
+description = "django-template-partials"
+optional = false
+python-versions = "*"
+groups = ["main"]
+files = [
+    {file = "django_template_partials-24.4-py2.py3-none-any.whl", hash = "sha256:ee59d3839385d7f648907c3fa8d5923fcd66cd8090f141fe2a1c338b917984e2"},
+    {file = "django_template_partials-24.4.tar.gz", hash = "sha256:25b67301470fc274ecc419e5e5fd4686a5020b1c038fd241a70eb087809034b6"},
+]
+
+[package.dependencies]
+Django = "*"
+
+[package.extras]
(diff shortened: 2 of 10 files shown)


phycodurus (commented):
cool and fun - I like this a lot. Should we add dynamic filtering on names in the same PR, so they release together? (because if I've got a zillion targets I'd probably like to filter them down so I don't have to scroll for too long). There may be additional considerations for extra large number of targets. Maybe we can discuss at stand-up...

Fingel:
> cool and fun - I like this a lot. Should we add dynamic filtering on names in the same PR, so they release together? (because if I've got a zillion targets I'd probably like to filter them down so I don't have to scroll for too long). There may be additional considerations for extra large number of targets. Maybe we can discuss at stand-up...

I think a follow-up PR. 

The layout of this page is so messed up. If you were to type to filter in the name, the input is all the way up in the top right next to the giant map, you can barely even see the table of data as is changes. I think we need to discuss a serious cleanup of this page. Like, what's up with the merge targets thing? It looks so haphazard.

tom_targets/templates/tom_targets/target_list.html
```diff
@@ -23,8 +23,7 @@
       </div>
     </div>
     {% select_target_js %}
-    {% aladin_skymap object_list %}
-    {% bootstrap_pagination page_obj extra=request.GET.urlencode %}
+    {% aladin_skymap targets %}
```
jchate6, line 26: This will only display the initial set of targets on the sky map. 
Without pagination, this won't update as new targets are added.

jchate6 (changes requested):
I feel like for all of it's flaws, pagination is still a better solution for getting to the end of large data sets. 
Also, pagination lets you change the targets on the skymap.

How hard would it be to only use infinite scrolling if there are fewer than 5 pages (100 objects) in a list? That would let people easily skip to the end of a table, and the sky chart should be able to handle 100 targets without too much delay, so if we aren't paginating we just show them all.

Thoughts? It just feels like until we have sorting available, this effectively locks the bottom of long tables away from the user.

Fingel:
> I feel like for all of it's flaws, pagination is still a better solution for getting to the end of large data sets. Also, pagination lets you change the targets on the skymap.
> 
> How hard would it be to only use infinite scrolling if there are fewer than 5 pages (100 objects) in a list? That would let people easily skip to the end of a table, and the sky chart should be able to handle 100 targets without too much delay, so if we aren't paginating we just show them all.
> 
> Thoughts? It just feels like until we have sorting available, this effectively locks the bottom of long tables away from the user.

You are right that with pagination it's easier to get to the end of large datasets, but the easier way to get to the end of a large dataset is with sorting. We do have sorting available - it's the ordering filter. There are options in there to sort by creation, update and name, both directions. It's not even clear to the user how the table is sorted by default, so I'm not sure why anyone would want to be checking the last page anyway.

Ideally we'd have sorting in the table headers instead of the filter, but we don't even display the creation/update dates in the table which I'd argue are arguable the most useful things to sort by.

tom_targets/templates/tom_targets/target_list.html
Fingel: I hate this map with a passion. It takes up too much space, it is by far the slowest part of the page, it keeps your mouse pointer hostage so you can't scroll the page like normal. I'm dubious of it's usefulness especially since now it only shows the objects in the current pagination set.

This map really needs to be fed via ajax as to not block the page, and ideally it shows all the targets that match the current filter, not just the first 25. For huge datasets some kind of sampling would be necessary.

jchate6 (approved):

Issue #1187, opened 2025-02-21
Aladin Skymap blocking load, catalog per-target.


Fingel:
**Describe the bug**
Currently the Aladin skymap works by parsing targets that have been directly rendered on the the target list page. While simple, this causes a few issues:

1. Targets need to be rendered on the page before the rest of the script is rendered, potentially increasing load time.
2. There is no way to dynamically update the points on the map, meaning a full page refresh is required. 

In particular <span>#</span>2 is problematic if changes to the target table (infinite scroll, filtering, sorting, etc) make it dynamic, as there is no way to update the targets on the Aladin map.

Another issue is that currently an entire Aladin "Catalog" is created per-target. From anecdotal evidence, this means that performance on the map begins to suffer with ~200 targets, becoming basically unusable with a few hundred more than that. The map should be capable of displaying many more targets. Ideally, we would show all targets that match the supplied filters, regardless if they are displayed in the current pagination page or not.

PR #1190, opened 2025-02-26
1187 aladin skymap blocking load catalog per target

Fingel:
Uses a dedicated JSON endpoint to load targets for the Aladin map, making it efficient enough that _all_ (well, up to 10k) targets that match a filter can be displayed on the map, instead of only the 25 in the current pagination page. This is much more intuitive and what I'd expect to see on the map.

The 10k number can be adjusted but at that point there's so many targets that it's not worth adding more as you won't really be able to see a difference.

This is important because with infinite scroll, it's not possible to keep what is displayed on the Aladin map consistent with the current "page". 

Fixes [#1187](https://github.com/TOMToolkit/tom_base/issues/1187)

https://github.com/user-attachments/assets/e7413d70-95bd-45b9-8cc7-e6aeb06cd706

### DIFF
--- a/tom_targets/views.py
+++ b/tom_targets/views.py
@@ -12,7 +12,7 @@ from django.core.management import call_command
 from django.db import transaction
 from django.db.models import Q
 from django_filters.views import FilterView
-from django.http import HttpResponse
+from django.http import HttpResponse, JsonResponse
 from django.http import HttpResponseRedirect, QueryDict, StreamingHttpResponse, HttpResponseBadRequest
 from django.forms import HiddenInput
 from django.shortcuts import redirect, render
@@ -98,6 +98,27 @@ class TargetListView(PermissionListMixin, FilterView):
         if self.request.htmx:
             return ['tom_targets/partials/target_table.html#target-table-inline']
         return super().get_template_names()
+
+
+class AladinTargetListView(PermissionListMixin, FilterView):
+    model = Target
+    filterset_class = TargetFilter
+    permission_required = f'{Target._meta.app_label}.view_target'
+    ordering = ['-created']
+
+    def render_to_response(self, context, **kwargs) -> HttpResponse:
+        qs = context['object_list'].only("name", "ra", "dec")
+        target_count = qs.count()
+
+        # If the queryset is too large, take a random sample
+        if target_count >= 10_000:
+            qs = qs.order_by('?')[:10_000]
+
+        targets = [
+            {'name': target.name, 'ra': target.ra, 'dec': target.dec}
+            for target in qs
+        ]
+        return JsonResponse({"targets": targets})
 
 
 class TargetNameSearchView(RedirectView):
--- a/tom_targets/filters.py
+++ b/tom_targets/filters.py
@@ -130,6 +130,13 @@ class TargetFilter(django_filters.FilterSet):
         }
     )
 
+    limit = django_filters.NumberFilter(label="Result Limit", method='filter_limit')
+
+    def filter_limit(self, queryset, name, value):
+        if value is not None:
+            return queryset[:value]
+        return queryset
+
     class Meta:
         model = Target
         fields = ['type', 'name', 'key', 'value', 'cone_search', 'targetlist__name']
--- a/tom_targets/templates/tom_targets/partials/aladin_skymap.html
+++ b/tom_targets/templates/tom_targets/partials/aladin_skymap.html
@@ -1,7 +1,25 @@
 <!-- embedding Aladin Lite example code found at: https://aladin.cds.unistra.fr/AladinLite/doc/  -->
-<div id="aladin-lite-div" style="width:auto; height:700px; margin:auto; margin-top: 20px;" ></div>
-
+{% load bootstrap4 tom_common_extras %}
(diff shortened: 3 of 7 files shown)


tom_targets/views.py
```diff
@@ -100,6 +100,27 @@ def get_template_names(self) -> list[str]:
...
+    permission_required = f'{Target._meta.app_label}.view_target'
+    ordering = ['-created']
+
+    def render_to_response(self, context, **kwargs) -> HttpResponse:
+        qs = context['object_list'].only("name", "ra", "dec")
+        target_count = qs.count()
+
+        # If the queryset is too large, take a random sample
+        if target_count >= 10_000:
+            qs = qs.order_by('?')[:10_000]
```
jchate6, line 115: I'd argue the most recently added would be better than random.

jchate6 (approved):
I have some misgivings about these changes. I feel like we are losing a fair amount of utility for the skymap for the sake of accommodating unnavigable lists of targets.

The ability to manipulate which individual targets are displayed could be very useful when there are only a few.

I understand the 10k is just an arbitrary choice, but it's tiny for a full TOM DB. Is there an actual upper limit where performance is noticeably affected? It feels like we shouldn't have a warning as the default status for the page most TOMs use as their defacto home page.

jchate6:
I'm happy to merge this into 1186 and see how users feel when everything is together.

Fingel:
> I have some misgivings about these changes. I feel like we are losing a fair amount of utility for the skymap for the sake of accommodating unnavigable lists of targets.
> 
> The ability to manipulate which individual targets are displayed could be very useful when there are only a few.
> 
> I understand the 10k is just an arbitrary choice, but it's tiny for a full TOM DB. Is there an actual upper limit where performance is noticeably affected? It feels like we shouldn't have a warning as the default status for the page most TOMs use as their defacto home page.

I'm super curious what functionality we are giving up? There are other (better) ways to hide individual markers, if that's what you need to do: for example, clicking one and choosing "hide".

I think it's strange that the skymap shows 25 targets currently, even if I have 1000's in my database. Why 25? Because that's the number we landed on for pagination? Now _that_ is arbitrary. 

In any place where a graph of objects like this appears, I think it's natural to assume that is represents all objects that _match your filters_, not just an arbitrarily small number of them that happen to map to what is currently shown in the table. Take a cone search for example. I want to find all objects within x arcseconds of some star. Wouldn't it be more natural to see a full representation of them, not just 25? I  don't care what page of results I happen to be shown, pages have nothing to do with the question.

10k is an arbitrary upper limit, the actual limit on the map is much higher. Fine with bumping up this number.

jchate6:
Good points. I think this is an issue of use cases. This display has different use cases based on circumstance and each TOM will use it differently, but our goal here is to make it flexible enough to meet those varied needs while having a sensible default functionality. 

The different use cases that I'm thinking of are as follows:

1. A default list of targets:
   - This could represent my entire DB or just some filtered/sorted subset of default "interesting" targets. 
   - displayed when I first arrive on the page.
   -  In either case displaying the FULL list on my plot is not practical. When I first go to this page, seeing the top N (definitely arbitrary from our perspective, but with n<100 ) targets on the list is really what I want. I want to know where in the sky the (north/south, proximity to the sun/moon) the most interesting targets on my list are at a glance. Hiding these targets among a sea of thousands of other targets makes this entire figure useless.
   - By default we assumed that the 25 most recently added targets are the most interesting, and that also fully represented the targets visible on the page when we were using pagination, and the displayed targets would update as the page changed. 

2. A filtered list:
    - A non-default query, showing some subset of targets determined by the filters.
    - Displayed after altering the filter values on the page.
    - Here I agree that if I've performed a specific query, we should return all the targets. I agree that if we do a cone search I should see all the targets in that cone. 

> I'm super curious what functionality we are giving up? There are other (better) ways to hide individual markers, if that's what you need to do: for example, clicking one and choosing "hide".

It's definitely flawed, but the ability to hide certain targets by name is definitely important functionality. I do not see a way to click on one and choose hide.
 
> I think it's strange that the skymap shows 25 targets currently, even if I have 1000's in my database. Why 25? Because that's the number we landed on for pagination? Now _that_ is arbitrary.

Agreed, but showing all of them effectively removes the utility of the plot. The idea was that displaying the targets on the page at least mapped the page to the plot so you could know exactly what was being displayed.

> 
> In any place where a graph of objects like this appears, I think it's natural to assume that is represents all objects that _match your filters_, not just an arbitrarily small number of them that happen to map to what is currently shown in the table. Take a cone search for example. I want to find all objects within x arcseconds of some star. Wouldn't it be more natural to see a full representation of them, not just 25? I don't care what page of results I happen to be shown, pages have nothing to do with the question.

Agreed. Unless my filter is `all`. 

> 
> 10k is an arbitrary upper limit, the actual limit on the map is much higher. Fine with bumping up this number.

Does it need to have a limit?  Can we give the user the option to display the top N results?

Do those points make sense?

Fingel:
I added a limit filter, which should alleviate many of these issues. For example, only displaying the most recent 25 targets, in both the map and the table, would be achieved by entering 25 into the result limit filter.

PR #1191, opened 2025-02-26
Only show target lists user has access to

### LINKED ISSUE #1141: Users should not be able to see Target Groups  on the Target List page that they don't have permission to access

Fingel:
This commit solves the issue where on the target list page, the target list dropdown for bulk operations shows all target lists on the TOM, not just the ones the user has access too.

Additionally, this commit adds the option to select groups for which a newly created target list belongs to, giving users of those groups access to it.

Fixes #1141

### DIFF
--- a/tom_targets/forms.py
+++ b/tom_targets/forms.py
@@ -271,7 +271,7 @@ class PersistentShareForm(AdminPersistentShareForm):
 
 
 class TargetListForm(forms.ModelForm):
-    groups = forms.ModelMultipleChoiceField(
+    user_groups = forms.ModelMultipleChoiceField(
         Group.objects.none(),
         required=False,
         widget=forms.CheckboxSelectMultiple,
@@ -280,4 +280,4 @@ class TargetListForm(forms.ModelForm):
 
     class Meta:
         model = TargetList
-        fields = ['name', 'groups']
+        fields = ['name', 'user_groups']
--- a/tom_targets/forms.py
+++ b/tom_targets/forms.py
@@ -268,3 +268,16 @@ class PersistentShareForm(AdminPersistentShareForm):
         if self.target_id:
             self.fields['target'].initial = self.target_id
             self.fields['target'].disabled = True
+
+
+class TargetListForm(forms.ModelForm):
+    groups = forms.ModelMultipleChoiceField(
+        Group.objects.none(),
+        required=False,
+        widget=forms.CheckboxSelectMultiple,
+        help_text='Select groups to share this target list with.'
+    )
+
+    class Meta:
+        model = TargetList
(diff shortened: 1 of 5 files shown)


tom_targets/forms.py
```diff
@@ -268,3 +268,16 @@ def __init__(self, *args, **kwargs):
         if self.target_id:
             self.fields['target'].initial = self.target_id
             self.fields['target'].disabled = True
+
+
+class TargetListForm(forms.ModelForm):
+    groups = forms.ModelMultipleChoiceField(
```
jchate6, line 274: How do we feel about renaming this field "User Groups"
I feel like there is a lack of clarity concerning the differentiation of Target Groups and User Groups both in the code and on the frontend.

tom_targets/views.py
```diff
@@ -776,12 +776,17 @@ class TargetGroupingCreateView(LoginRequiredMixin, CreateView):
...
 
+    def get_form(self, *args, **kwargs):
+        form = super().get_form(*args, **kwargs)
+        form.fields['groups'].queryset = self.request.user.groups.all()
+        return form
+
     def form_valid(self, form):
         """
-        Runs after form validation. Saves the target group and assigns the user's permissions to the group.
+        Runs after form validation. Saves the target group and assigns the user and group permissions to the group.
```
jchate6, line 789: Can we make this comment:
`Runs after form validation. Saves the target list and assigns the user and group permissions to the list.`

jchate6 (changes requested):
Just a few tweaks to clarify which "group" we are referring to at any given moment.

tom_targets/views.py
```diff
@@ -791,6 +796,11 @@ def form_valid(self, form):
         assign_perm('tom_targets.view_targetlist', self.request.user, obj)
         assign_perm('tom_targets.change_targetlist', self.request.user, obj)
         assign_perm('tom_targets.delete_targetlist', self.request.user, obj)
+
+        for group in form.cleaned_data['groups']:
```
jchate6, line 800: This will also need to be updated.

Fingel:
I did some renaming. This doesn't help the fact that the UI still stays target grouping everywhere, for example on the list page: /targets/targetgrouping/

Not sure if that something we want to tackle in this PR or not.

jchate6:
We actually have an issue for handling this generally:
 - #679 
 
 I think we leave it for now.

jchate6 (approved):

PR #1196, opened 2025-03-05
Silence log output in tests. Fix naive datetimes.

Fingel:
Prior to this commit the output of running the test suite was filled with many, many helpful lines such as "test user has no profile, creating profile" and "Target Post Save Hook". This was not only annoying but also caused 3rd party tools to interpret the test output as erroneous. See https://github.com/TOMToolkit/tom_base/actions/runs/13683264805

This commit also fixes a few places where we were using naive datetimes which was the cause of much of the noise in the test output as well.

image


jchate6 (approved):
nice. Thanks for this.

PR #1197, opened 2025-03-06
Limit alert query results to 100

Fingel:
Alternatively, we could just not show the result table at all.
image


jchate6 (approved):
I like this.
I'd imagine that if someone returns more than 100 results they wouldn't really be able to sort through them to find the couple they wanted without re-doing the query.

jchate6:
@rachel3834 I wanted to get your opinion on this. 
If there are more than 100 results from a query, how do we feel about only displaying the 1st 100?

Hopefully this is a temporary solution and we can fix the problem permanently in the future, but in the meantime, is asking the user to constrain their query reasonable?

PR #1199, opened 2025-03-13
1194 async tasks django

Fingel:
Note, once this PR is approved I will remove the demo code. This is here for demonstration purposes.


phycodurus (approved):
Nice! I wonder if at some point we'll just install `django-tasks` into `tom_base` by default, but that's probably out of scope here.

Fingel:
> Nice! I wonder if at some point we'll just install `django-tasks` into `tom_base` by default, but that's probably out of scope here.

This MR does add django_tasks as a dependency of tom_base, as dramatiq is a dependency currently.

phycodurus:
> > Nice! I wonder if at some point we'll just install `django-tasks` into `tom_base` by default, but that's probably out of scope here.
> 
> This MR does add django_tasks as a dependency of tom_base, as dramatiq is a dependency currently.

Opps. I was mistakenly think that was as change in `tom_setup`.

jchate6:
Removing myself as reviewer so I don't hold this PR up. 
Let me know if you really think I should have an opinion.

Fingel:
I've removed the demo code. I think this should be good now - I've confirmed that existing toms without django-tasks added to INSTALLED_APPS can still execute the atlas code synchronously. Adding django_tasks to INSTALLED_APPS is only necessary for enabling the async database backend. So this PR should be backwards compatible.

The only issue that should arise is for toms that are already using dramatiq. In that case the tom owner can still keep using dramatiq - they will need to make sure they include it as an explicit dependency of their tom. Add this to release notes?

**This release adds django-tasks as the default task backend. For more information, see DEP 0014: https://github.com/django/deps/blob/main/accepted/0014-background-workers.rst. If you are currently utilizing dramatiq as a task backend for your TOM, please make dramatiq an explicit dependency in requirements.txt or pyproject.toml**

PR #1201, opened 2025-03-18
move DRF create_auth_token signal receiver to signals

phycodurus:
refactors signal receiver out of models.py to signals.py consistent with Django docs recommendation. See
https://docs.djangoproject.com/en/5.1/topics/signals/


jchate6 (commented):
looks good

PR #1207, opened 2025-04-08
Improve observation list page

griffin-h:
Our observation list was not working because of the large number of targets in our database. The view would try to populate the `ModelMultipleChoiceFilter` with targets and then the page would never load. This PR restricts that filter only to targets with observations, which is probably more useful even for users with smaller databases.

While I was at it, I added choices for the facility and status, based on existing possible values.

Once the page was loading, the table was too big to fit in its `col-md-10`, so I removed the "end" column and the lists of aliases.

Two other fixes to small bugs:
- the `scheduled_end` filter was not appearing
- the pagination was not working


jchate6 (approved):
This PR seems fine. I can't replicate the failing tests. Investigating...

jchate6:
I'm going to approve this and merge it into another branch (other than dev) for testing...

PR #1208, opened 2025-04-08
Bug/observation tests broken

jchate6:
Griffin's PR was breaking tests that I couldn't get to fail locally. Pulling into here to figure out why.


jchate6:
@griffin-h sorry for any confusion about your PR. 
Running tests locally passed, because I was still pointing to my default DB before the test TB was created, but this code was being run while it was compiled instead of during the tests themselves, so without a proper DB it was breaking.
I've moved it into the init like we do for forms that need updatable choice fields.

tom_observations/views.py
```diff
@@ -46,10 +46,23 @@ class ObservationFilter(FilterSet):
...
+        queryset=Target.objects.filter(observationrecord__isnull=False).distinct().order_by('name')
+    )
     observationgroup = ModelMultipleChoiceFilter(
         label='Observation Groups', queryset=ObservationGroup.objects.all()
     )
+    facility = MultipleChoiceFilter(choices=facility_choices())
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.status = MultipleChoiceFilter(
```
phycodurus, line 60: How about adding a comment that explains that
```
# status filter is populated dynamically via the list comprehension here in
# the __init__ (at runtime) when the ObservationRecord database table exists
# (and the db table doesn't necessarily exist at Class-interpretation-time)
```

phycodurus (approved):

PR #1211, opened 2025-04-17
add customization clarification and links to tom_demoapp

jchate6:
Please take a look at the tom_demoapp wiki as well.
https://github.com/TOMToolkit/tom_demoapp/wiki


phycodurus (approved):

phycodurus:
This may be the topic of a separate issue:

At some point, I think it will be beneficial to define in `tom_base`, a
`class TOMToolkitAppConfig(AppConfig)` that `tom_demoapp` and most other TOMToolkit INSTALLED_APPs should sublcass. The `TOMToolkitAppConfig` class would provide stubs for the integration points that documented here. This would reduce the boilerplate code that a new TOMToolkit INSTALLED_APP would need to write.

PR #1225, opened 2025-05-27
Feature/profile cipher for securing data

phycodurus:
This PR covers the signals, utility functions, and other changes required to store sensitive data, encrypted in the TOMToolkit database. See [tom_eso](https://github.com/TOMToolkit/tom_eso/pull/20) for an example of how to use it. Briefly,

- a new UserSession model is create for each User upon log in. It just has two foreign keys to the User and the User's Session. It allows access to a User's Session outside the context of a request, where the Session is added by the middlewhere.
- When a user changes their password and the Session is updated to keep them logged in, the `UserSession` is updated in `tom_base.view.UserUpdateView`.
- A set of signal receivers in `signals.py` looks for saves to the User model and calls out to `session_utils.py` to keep the EncryptedBinaryFields up-to-date.
- Some tests have been added in `tom_common.test.py`
-


tom_common/tests.py
```diff
@@ -308,3 +321,272 @@ def test_nonexistent_custom_robots_txt(self):
...
+        # TODO: make sure these type-hints are accurate!!!!
+        # why are calling extract_key_from_session_store with what looks
+        # like a Session object? Is it really a Session instance or is it really a
+        # SessionStore instance?
+
+        # here, we are calling extract_key_from_session_store with a Session instance
+        weird_encryption_key: bytes = extract_key_from_session_store(session)
+
+        # Extract the encryption key from the session.
+        encryption_key: bytes = extract_key_from_session(session)
```
Fingel, line 444: It's confusing why there is both `extract_key_from_session_store` and `extract_key_from_session` but one of them is weird? Shouldn't there be just a single function that's not weird? Which one are you supposed to use?

jchate6 (commented):
I'm confused by the `reencrypt_app_fields` integration point.

Is there a benefit to using `auto_reencrypt_model_instances_for_user` rather than just passing the field name and keeping the rest of the code in tom_base?

This kinda goes against all of the other integration points, where we pass a list of dictionaries. That's not necessarily a problem, I'm just wondering if we would anticipate users doing anything else to modify this code (changing ciphers, something else?) and if not, should we just pass the field string instead?

tom_common/tests.py
phycodurus: Yeah, that's confusing. Here's how I explained it in the message
for the commit I just made to clean some of this up:

> The `SessionStore` is the 'normal' way to access the data in the
> `Session` model via the dictionary-like API it provides. And this
> is how we do it when we have a `HTTPRequest` instance that has been
> decorated with a `SessionStore` object in the middleware. However,
> when we need access without a decorated `HTTPRequest`, we use the
> `UserSession` which gives us access (it has a ForeignKey) to the
> underlying `Session` model. Then we use session.get_decode().

So, we use get_key_from_session_store when we can (have HTTPRequest) and get_key_from_session when we have to (only have UserSession).

jchate6 (changes requested):
This breaks if the user password is changed by admin via the `tom_common.views.UserPasswordChangeView` route. I assume this has something to do with either not catching the profile update signal and/or not being in the correct session with the correct user when the password is changed?

tom_common/tests.py
```diff
@@ -308,3 +321,272 @@ def test_nonexistent_custom_robots_txt(self):
...
+        self.assertIsInstance(encryption_key, bytes)
+
+        self.assertEqual(encryption_key, weird_encryption_key)  # check that they are the same
+        # if they are the same, that's weird
```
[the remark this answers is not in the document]
phycodurus: I made a new test to test that `get_key_from_session_model()` and `get_key_from_session_store()` return the same key. That test was buried in the test that the encryption_key is updated upon password change (and it definitely didn't belong there).
I hope the doc string for the new test also helps clarify things (it's pretty much the same as the comment above).

jchate6 (changes requested):
This does not seem to work with multiple encrypted fields in a model. 

i.e. having
`external_password`
`external_api_key`

as part of your profile results in both values being synced unless I set something up wrong.

phycodurus:
> This breaks if the user password is changed by admin via the `tom_common.views.UserPasswordChangeView` route. I assume this has something to do with either not catching the profile update signal and/or not being in the correct session with the correct user when the password is changed?
> 
This has been address in the latest commits. We now clear a User's encrypted fields if and Admin changes their password. Upon that password change, the data in those fields became irretrievable and stale. So, it makes sense to clear it out.

Also, added a confirmation and warning to the admin that encrypted fields for the User would be cleared out.

tom_common/templates/auth/user_confirm_change_password.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'tom_common/base.html' %}
+{% block content %}
+{% block title %}<h4>Confirm password change</h4>{% endblock %}
+<form class="form" method="post">
+  {% csrf_token %}
+  <p>Are you sure you want to change the password for user: <b>{{ object.username }}</b>?</p>
+  <p>This will also clear out all of their saved external service API keys and passwords (if any).</p>
```
jchate6, line 7: Can we put an if statement here to make sure that `object.username` doesn't match the current username?
This (correctly) is not true in that case.

tom_common/models.py
```diff
@@ -9,3 +16,138 @@ class Profile(models.Model):
...
+            raise AttributeError(
+                f"A Fernet cipher must be set on the '{owner.__name__}' instance "
+                f"as '_cipher' to access property '{self.property_name}'."
+            )
+
+        encrypted_value = getattr(instance, self.db_field_name)
+        if not encrypted_value:
+            return ''
+
+        # Handle bytes vs memoryview
```
jchate6, line 75: This is not a great comment.
Can you explain WHY we need to handle bytes vs memoryview?

tom_common/models.py
```diff
@@ -9,3 +16,138 @@ class Profile(models.Model):
...
+    """
+    # if either of the referenced objects are deleted, delete this object (CASCADE).
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    session = models.ForeignKey(Session, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return f'UserSession for {self.user.username} with Session key {self.session.session_key}'
+
+
+class EncryptedProperty:
```
jchate6, line 39: Is it not possible to do the whole
```session: Session = UserSession.objects.get(user=user).session
session_store: SessionStore = SessionStore(session_key=session.session_key)
cipher_key: bytes = get_key_from_session_store(session_store)
cipher = Fernet(cipher_key)
user.testprofile._cipher = cipher
```
dance here? or in session_utils?
This seems like a lot of boiler plate that is required for each use of this code...

tom_common/models.py
```diff
@@ -9,3 +16,138 @@ class Profile(models.Model):
...
+    """
+
+    def reencrypt_model_fields(self, decoding_cipher: Fernet, encoding_cipher: Fernet) -> None:
+        """Re-encrypts all fields managed by an EncryptedProperty descriptor."""
+        model_save_needed = False
+        for attr_name in dir(self.__class__):
+            attr = getattr(self.__class__, attr_name)
+            if isinstance(attr, EncryptedProperty):
+                try:
+                    # Set decoding cipher and get plaintext
```
jchate6, line 112: Again, I don't feel like comments like these are useful. I can see what the code does. If I'm trying to debug this in 6 months, comments about WHY we need to pull out the plaintext are much more helpful.

tom_common/session_utils.py
```diff
@@ -0,0 +1,267 @@
...
+    # The key was stored as a string, so we encode it back to bytes.
+    return key_as_str.encode('utf-8')
+
+
+def get_key_from_session_store(session_store: SessionStore) -> bytes:
+    """Extracts the encryption key from a Django SessionStore instance.
+
+    This function first retrieves the underlying Session object using the
+    session_key from the SessionStore. It then calls
+    `extract_key_from_session` to get the actual key.
```
jchate6, line 106: I don't think this is true anymore

tom_common/session_utils.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def reencrypt_data(user) -> None:
+    """Re-encrypts sensitive data for a user after a password change.
+
+    This function is typically called when a user changes their password. It performs
+    the following steps:
+    1. Retrieves the current encryption key from the user's session.
+    2. Creates a new encryption key based on the user's new password.
+    3. Iterates through all installed Django apps, calling a `reencrypt_app_fields`
```
jchate6, line 129: I think this is outdated now... `reencrypt_encypted_fields_for_user`?

jchate6 (changes requested):
Sorry, ran out of time. Will finish the review on the rest of the code next week, but didn't want to leave you hanging for tomorrow!

tom_common/models.py
phycodurus: Good point. Here's why (I added this comment to the code:
```python
            # postgresql/psycopg uses a memoryview object for BinaryFields.
            # Sqlite3 uses bytes. When needed, convert to the encrypted_value
            # to bytes before we decrypt and decode it.
```

tom_common/session_utils.py
phycodurus: True. Doc-string updated.

tom_common/models.py
phycodurus: Yes! Two new helper functions have been added to `session_utils.py`:
  - `set_encrypted_field(user, model_instance, field_name, value)`
  - `get_encrypted_field(user, model_instance, field_name)`

and that does really simplify things.

tom_common/session_utils.py
```diff
@@ -0,0 +1,362 @@
...
+    from django.contrib.auth.models import User  # noqa
+    from tom_common.models import EncryptableModelMixin
+
+    for model_class in app_config.get_models():
+        if issubclass(model_class, EncryptableModelMixin):
+            logger.debug(f"Found EncryptableModelMixin subclass: {model_class.__name__} in app {app_config.name}")
+            # the Model must have a ForeignKey to the User (and AppProfile Models do)
+            if hasattr(model_class, user_relation_field_name):
+                try:
+                    # Handles OneToOneField or unique ForeignKey to User
```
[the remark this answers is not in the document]
jchate6: I don't think you need to differentiate this from the unlikely multiple instances below, but if you want to, maybe do so more explicitly `# in the case of a single instance per user:` or similar

tom_common/session_utils.py
```diff
@@ -0,0 +1,362 @@
...
+            else:
+                logger.warning(f"Model {model_class.__name__} is Encryptable but does not have a direct "
+                               f"'{user_relation_field_name}' attribute. Cannot automatically fetch instance "
+                               f"for user {user.username} for re-encryption.")
+
+
+def clear_encrypted_fields_for_user(
+    app_config: AppConfig,
+    user: 'User',
+    user_relation_field_name: str = 'user'
```
jchate6, line 325: This will also break if I do not use `user` for my profile user OneToOneField

tom_common/session_utils.py
```diff
@@ -0,0 +1,362 @@
...
+    """
+    from tom_common.models import EncryptableModelMixin
+
+    for model_class in app_config.get_models():
+        if issubclass(model_class, EncryptableModelMixin):
+            logger.debug(f"Found EncryptableModelMixin subclass: {model_class.__name__} in "
+                         f"app {app_config.name} for clearing.")
+            if hasattr(model_class, user_relation_field_name):
+                try:
+                    # Handles OneToOneField or unique ForeignKey to User
```
jchate6, line 346: Same comments as above.

tom_common/session_utils.py
```diff
@@ -0,0 +1,362 @@
...
+        user: The User object associated with the encrypted data.
+        instance: The model instance containing the EncryptedProperty.
+        field_name: The string name of the EncryptedProperty to access.
+
+    Returns:
+        The decrypted string value, or None if decryption fails for any reason
+        (e.g., no active session, key not found).
+    """
+    try:
+        session: Session = UserSession.objects.get(user=user).session
```
jchate6, line 148: This breaks with a `tom_common.models.UserSession.MultipleObjectsReturned: get() returned more than one UserSession -- it returned 2!` if the user is logged into two or more Sessions.

Should probably not be a "get"

tom_common/session_utils.py
```diff
@@ -0,0 +1,362 @@
...
+        user: The User object associated with the encrypted data.
+        instance: The model instance containing the EncryptedProperty.
+        field_name: The string name of the EncryptedProperty to set.
+        value: The plaintext string value to encrypt and set.
+
+    Returns:
+        True if the field was set successfully, False otherwise.
+    """
+    try:
+        session: Session = UserSession.objects.get(user=user).session
```
jchate6, line 194: Same issue as above with multiple sessions.

jchate6 (changes requested):
A few more things

tom_common/views.py
```diff
@@ -165,23 +201,72 @@ def dispatch(self, *args, **kwargs):
...
+            # The old UserSession (if any) linked to the old Session would have been deleted by CASCADE.
+            # We need to create a new UserSession linking the User to the new Session.
+            new_session_key = self.request.session.session_key
+            if new_session_key:
+                try:
+                    # Get the new Django Session object from the database
+                    # Need to import Session model: from django.contrib.sessions.models import Session
+                    from django.contrib.sessions.models import Session
+
+                    new_session = Session.objects.get(session_key=new_session_key)
```
jchate6, line 243: Multiple session issues

tom_common/signals.py
```diff
@@ -1,33 +1,168 @@
...
+    else:
+        logger.debug(f"UserSession already exists: {user_session}")
+
+
+# Signal: Delete UserSession on logout
+@receiver(user_logged_out)
+def delete_user_session_on_user_logged_out(sender, request, user, **kwargs) -> None:
+    """Whenever a user logs out, delete all their UserSession instances.
+    """
+    user_sessions = UserSession.objects.filter(user=user)
```
jchate6, line 88: This could cause issues if logged into other browsers.

jchate6: Seems to work actually. Loging out of one browser logs you out of all browsers

tom_common/session_utils.py
```diff
@@ -0,0 +1,357 @@
...
+    Automatically finds models in the app_config that inherit from EncryptableModelMixin
+    and attempts to re-encrypt their fields for the given user.
+
+    :param app_config: The AppConfig instance of the plugin app.
+    :param user: The User whose data needs re-encryption.
+    :param decoding_cipher: Fernet cipher to decrypt existing data.
+    :param encoding_cipher: Fernet cipher to encrypt new data.
+    """
+    # Import models here, when the function is called, ensuring apps are ready.
+    from django.contrib.auth.models import User  # noqa
```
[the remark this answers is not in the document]
phycodurus: The problem with this one was that when it was at the top of the file (as usual),  when Django was loading, sometimes the import would happen before the `django.contrib.auth` app was `ready()`. So, the import was moved here to happen at runtime when the `auth` app was surely ready. (But that might have been before some of the refactoring). I'll try moving it back to normal and do some testing....

tom_common/session_utils.py
```diff
@@ -0,0 +1,357 @@
...
+    Finds models in an app that are Encryptable and clears their encrypted fields for the given user.
+
+    This is a destructive operation used when a user's password is reset without
+    them being logged in, making the old decryption key unavailable. This happens,
+    for example, when an adminitrator resets their password.
+
+    :param app_config: The AppConfig instance of the plugin app.
+    :param user: The User whose data needs to be cleared.
+    """
+    from tom_common.models import EncryptableModelMixin
```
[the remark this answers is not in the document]
phycodurus: true. address in latest commit

PR #1253, opened 2025-07-09
Clean orphaned permissions after target delete

### LINKED ISSUE #1249: Group object permissions are not deleted when the related object is deleted
**Describe the bug**
I have entries in the `guardian_groupobjectpermission` table that reference rows in other tables that no longer exist. I'm assuming these should be deleted in a cascade when the parent object is deleted.

**To Reproduce**
1. Create a private target with some group permissions. Note the target ID.
2. Delete that target through the GUI
3. In the database, run `select * from guardian_groupobjectpermission where object_pk='#####'`, inserting the ID of the deleted target in quotes.
4. See that there are rows returned

**Expected behavior**
I would expect these rows to be deleted when the target is deleted.

**Screenshots**

- OS: macOS
- Browser: Firefox
- Version: 140.0.2
- Python Version: 3.10

**Additional context**
This is not a major issue, but it would be nice to keep the permissions table from filling up with junk for large databases.

Fingel:
Cleans all orphaned Django-Guardian permissions after a Target is deleted. This makes sense to do here as Targets can have row level (private) permissions. We could potentially run this as well for dataproducts, though I'm less versed on how it works there. Either way, there's no reason to keep these objects around for any model, so it's fine to delete permissions that aren't associated with Targets.

Alternatively, we don't do any cleanup and leave it to the TOM developer to run the management command manually:
https://django-guardian.readthedocs.io/en/stable/api/management/ which is also fine.

Fixes #1249


jchate6 (approved):
nice solution. 
It's probably good to do this maintenance periodically anyway.

PR #1267, opened 2025-07-31
add ReducedDatum filters

### LINKED ISSUE #1125: Add `Target` and `DataProduct` filters to the  ReducedDatums DRF API.
I have some software used for fitting models to timeseries data.  I'd like this code to be able to directly retrieve from my TOM all lightcurves for a give target (identified by name).  

I can see the TOM already provides APIs for querying targets, and for listing, creating (uploading) and deleting data products (tom_dataproducts/api_views) but it is not clear to me whether downloading the ReducedDatums is supported.  

Please can you point me in the right direction?

jchate6:
This adds filters to the reduced Datums API
image

Are there other filters we should add?
These are all programatically accessible via a get request and can also be run from the UI url (i.e. `http://my_tom_url/api/reduceddatums/?data_product_pk=2` ).

TODO:
- [x] Tests
- [x] Docs


tom_dataproducts/filters.py
```diff
@@ -14,3 +22,29 @@ class Meta:
...
+        - data_product_pk: Filter by the primary key of the associated DataProduct.
+        - data_product_product_id: Filter by the "Product ID" or filename of the associated DataProduct.
+        - source_name: Filter by the name of the source.
+        - data_type: Filter by the type of data (e.g., 'photometry', 'spectrum').
+
+    Access these filters via the API endpoint:
+        `GET /api/reduceddatums/?target__id=<id>&data_type=<type>`
+    """
+    target_name = django_filters.CharFilter(label='Target Name', method='filter_name')
+    data_product_pk = django_filters.NumberFilter(field_name='data_product__pk', label='Data Product Primary Key')
```
Fingel, line 41: Small nitpick: While technically the PKs are integers, that's more of an implementation detail. Unless you plan on doing arithmetic with the PKs, almost always better to treat them as dumb strings. I highly doubt we'd ever change the pks to uuids or whatever, but just something to keep in mind.

Fingel (approved):
Looks good!

tom_dataproducts/filters.py
jchate6: I changed some things around a little, so I don't need to specify the type anymore. The type will be determined by the model field now.

Fingel: Perfect!

tom_observations/views.py
```diff
@@ -11,8 +11,8 @@
 from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.management import call_command
-from django_filters import (CharFilter, ChoiceFilter, DateTimeFromToRangeFilter, FilterSet, ModelMultipleChoiceFilter,
-                            OrderingFilter, MultipleChoiceFilter)
+from django_filters import CharFilter, ChoiceFilter, DateTimeFromToRangeFilter, ModelMultipleChoiceFilter, \
```
phycodurus, line 14: From [PEP8](https://peps.python.org/pep-0008/)
> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

phycodurus (approved):

KKruszynska (approved):
Looks good to me.

PR #1269, opened 2025-08-06
Feature/pass user from view to facilities

phycodurus:
This enhances the ObservationCreateView by

- Passes User instance to the Facility so it can use that to get data out of the Facility Profile. (Facility apps will know if they have a User profile or not)
- Also puts the user in the kwargs of the Form class for the same reason.
- Clean up instantiation itself so it only happens once per Request/Response cycle. This simplifies the changes above. What this means is that rather than calling `self.get_facility_class()()` all over the place, we call it once at the beginning (in `dispatch()`) and refer to `self.facility_instance` throughout the View.

These changes are prerequisites for

- [ ] https://github.com/TOMToolkit/tom_eso/issues/15

### DIFF
--- a/docs/api/tom_observations/facilities.rst
+++ b/docs/api/tom_observations/facilities.rst
@@ -38,3 +38,15 @@ Blanco
 
 .. automodule:: tom_observations.facilities.blanco
     :members:
+
+
+********************************
+Facilities as ``INSTALLED_APPS``
+********************************
+
+The following Facilities are implemented as ``INSTALLED_APPS``. As such,
+each resides in its own git repository:
+
+* `Neil Gehrels Swift Observatory (tom_swift) <https://github.com/TOMToolkit/tom_swift>`__
+* `European Southern Observatory (tom_eso) <https://github.com/TOMToolkit/tom_eso>`__
+* `Liverpool Telescope (tom_lt) <https://github.com/TOMToolkit/tom_lt>`__


tom_observations/views.py
```diff
@@ -144,6 +144,22 @@ class ObservationCreateView(LoginRequiredMixin, FormView):
     """
     template_name = 'tom_observations/observation_form.html'
 
+    def dispatch(self, request, *args, **kwargs):
+        """Figure out what HTTP method (GET, POST, etc) should be called
+        to handle this request.
+
+        Here, we extend the method to attach the Facility class to the View so
+        we don't have to create more and one instances of it. So,
```
jchate6, line 152: * more than one

tom_observations/views.py
```diff
@@ -312,8 +338,41 @@ def get_initial(self):
...
+
+        Here, we extend the super-class method to add the User instance:
+        call the super() and then add the user to the form kwargs,
+        here, in the View where, we can get the User instance from the request.
+        """
+        kwargs = super().get_form_kwargs()
+        kwargs['user'] = self.request.user
+        return kwargs
+
     def post(self, request, *args, **kwargs):
```
jchate6, line 352: I don't care for this as a method for handling this kind of error. 
This feels like something we should be handling at the base form level or the `get_form()` method  rather than trying to catch the exception here.

jchate6 (changes requested):
minor things...

tom_observations/views.py
phycodurus: This isn't a error handling change. This is where we attach the User instance to the form kwargs. We do that here b/c the View has the request and the request has the user.

jchate6 (changes requested):

Issue #1111, opened 2024-11-08
Build htmx table widget


jchate6:

phycodurus:
If we're zeroing in on Name search/sort/filter: I think dynamic name search is a pretty standard htmx use-case. The performance on large Target tables might be an issue.

(by "dynamic name search" I mean: type into a text box and have the displayed targets update accordingly. So, maybe I'm talking about filtering).

PR #1279, opened 2025-08-21
Feature/dynamic target table columns

### LINKED ISSUE #1278: Dynamic target list table columns
This is a sub issue of #1111 as that issue seems to presume that there is actually data in the Target table that would be desirable to search and sort on.

This PR introduces a new setting, `TARGET_LIST_COLUMNS` that determines which columns are displayed in the target list table. The default value is the same as what is displayed there now.

The implementation uses some custom tags to read the setting and read the corresponding values on a target by target basis. Some headers are treated specially, such as name (returns all names), observations and saved_data (are relational counts). Standard model fields will be displayed, tags and extra_fields are also supported.

This should enable TOM users an easy way to display the data they want for their particular TOM, without needing to write a custom table implementation.

Fingel:
Allows target list table to be configured via a setting specifying columns. This should enable users to display the attributes that are important to their project without the need for overriding templates or writing other custom code. This also means #1111 becomes more realistic as it broadens the fields that can be searched and sorted. 

Screenshot From 2025-08-21 10-23-20

closes #1278


tom_base/settings.py
```diff
@@ -311,6 +311,13 @@
 # Default permission for newly created targets. Values can be 'PRIVATE', 'PUBLIC', or 'OPEN'
 TARGET_DEFAULT_PERMISSION = 'PRIVATE'
 
+# Display these columns in the target list table. Values can be attributes or properties on
+# the Target model, tags or extra fields. The fields `observations` and `saved_data` are
+# special cases with custom implementation.
+TARGET_LIST_COLUMNS = [
```
phycodurus, line 317: This may be out-of-scope for this issue: We could make this user-specific rather than TOM-specific. Now that we have User Profiles, this could be part of the `tom_common.models.Profile` and a User could set their own fields (or fallback on a default set).

I suppose a generalization of this would be to allow the column specification to be a reference to callable that takes a Target instance and returns a value (possibly dynamic like "next_night_visible") and we put that in the value in a column for that Target's row).

Fingel: User configurable would be a good next step. It would really unlock being able to have the workflow where you can choose which columns are present directly from the page itself. Still I think project wide defaults are useful. For example, we display target type SIDEREAL or NON-SIDEREAL, even though I image most toms do not host targets of both types. So how useful is it? At the same time, it's very hard to come up with sensible defaults for all TOMS, so we're left with these least common denominators, not very good for anyone.

PR #1284, opened 2025-08-26
update update_observation_status

jchate6:
Handle multiple targets per observation.
Adds test.


tom_observations/facility.py
```diff
@@ -378,15 +378,15 @@ class BaseRoboticObservationFacility(BaseObservationFacility):
 
     def update_observation_status(self, observation_id):
```
phycodurus, line 379: I think a docstring explaining the use case that this supports is in order. It should answer the questions "why might we have more than one observation record with that same observation_id?, etc. (or reference the issue).

tom_observations/facility.py
```diff
@@ -378,15 +378,15 @@ class BaseRoboticObservationFacility(BaseObservationFacility):
...
         from tom_observations.models import ObservationRecord
-        try:
-            record = ObservationRecord.objects.get(observation_id=observation_id)
-            status = self.get_observation_status(observation_id)
+        records = ObservationRecord.objects.filter(observation_id=observation_id)
+        if not records:
+            raise Exception('No record exists for that observation id')
```
phycodurus, line 383: Should be:

`Exception(f'No records exist for observation id: {observation_id}')`

tom_observations/facility.py
```diff
@@ -378,15 +378,15 @@ class BaseRoboticObservationFacility(BaseObservationFacility):
...
+        status = self.get_observation_status(observation_id)
```
phycodurus, line 384: get_observation_status is an @abstract method and hopefully every facility subclass implementation can handle multiple ObservationRecords with the same observation_id. (and return a single status)

phycodurus (approved):

PR #1296, opened 2025-09-03
remove Comments header when comments are disabled

griffin-h:
Right now if you disable comments in the settings, there is still a header at the bottom of the page but with nothing underneath it. This just removes that header as well.


tom_targets/templates/tom_targets/target_detail.html
```diff
@@ -115,16 +115,15 @@ <h4>Observations</h4>
         {% spectroscopy_for_target target %}
       </div>
 
-      {% comments_enabled as comments_are_enabled %}
```
jchate6, line 118: You need this line to actually make the reference to `comments_are_enabled` valid.

`comments_enabled` is the tag defined in `tom_common_extras`  and we need to store the results in the `comments_are_enabled` variable.

As is, this just effectively disables comments for everyone.

jchate6 (changes requested):
small change.

PR #1297, opened 2025-09-03
check for duplicate reduceddatums during target merge

griffin-h:
Addresses the `TODO` in the `target_merge` function to handle duplicate `ReducedDatum` objects. If two targets are merged that have one or more identical `ReducedDatum` objects, the duplicates are deleted instead of being associated with the primary target.


jchate6 (approved):
Thanks!

PR #1298, opened 2025-09-04
Add time inputs to start/end windows on LCO observation form

Fingel:
This required some serious Django Form gymnastics. I totally understand if this is tech debt we don't want to take on. The Django Form API is not the greatest achievement in Python history.

Important thing here is that the time field defaults to midnight so that people that don't care about the exact time don't need to fill it in.

Screenshot From 2025-09-03 18-54-28


jchate6 (approved):
looks good

Issue #1303, opened 2025-09-06
Add a Basic General Facility Model


jchate6:
We want to add a DB model for a generic facility so users can add facilities without needing a full facility module.

For the bare bones General facility we need to develop:

- [ ] `tom_observations.models.Facility()` -- Establish the DB model with basic Data
  - [ ] site_code
  - [ ] mpc_site_code
  - [ ] full_name
  - [ ] short_name
  - [ ] latitude
  - [ ] longitude
  - [ ] elevation
  - [ ] diameter
  - [ ] typical_seeing
  - [ ] telescope_type (optical, infrared, radio, satellite, occultation)
  - [ ] info_url
  - [ ] api_url
- [ ] `tom_observaitons.admin.FacilityAdmin()` -- Allow admin to add, edit, remove facilities
- [ ] 'tom_observations.serializers.FacilitySerializer()` -- Allow programaitc access to facilities

rachel3834:
After thinking over the telescope type and location-related fields, I added fields to support both ground and space-based facilities.  I also generalised the telescope type field to be a general descriptor of the wavelength range or messenger the telescope is sensitive to.  Initially this seemed like it should be a choice field, but a number of telescopes cover multiple wavelength options, e.g. UV and optical and NIR.  So in the end I left this as a short charfield.

rachel3834:
After discussion, we decided in fact detector_type should be a choice field, so I've converted it back.

PR #1308, opened 2025-09-14
Addition of basic model for a general-purpose telescope facility

rachel3834:
Responding to issue #1303 

This PR includes the basic Facility model class plus associated admin and serializer functions.


tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+
+
+class Facility(models.Model):
+    """
+    Class representing individual telescope facilities.
+
+    :param site_code: Short-hand code used to reference the facility
+    :type site_code: str
+
+    :param mpc_site_code: Three-digit reference code for the facility from the Minor Planet Center
```
jchate6, line 195: Minor thing (I know I wrote this in the initial issue) the MPC technically refers to these as "Observatory Codes" We might want to change the name to correspond.

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+        ("leo", "Low-Earth Orbit"),
+        ("l2", "Lagrange Point 2"),
+        ("earth-trailing", "Earth-trailing orbit"),
+        ("geosync", "Geosynchronous orbit"),
+        ("polar", "Polar orbit"),
+        ("elliptical", "Elliptical orbit")
+    ]
+
+    site_code = models.CharField(max_length=20, null=True, blank=True)
+    mpc_site_code = models.CharField(max_length=3, null=True, blank=True)
```
jchate6, line 248: I would add `help_text="<a href='https://minorplanetcenter.net/iau/lists/ObsCodesF.html'>"
                                               "MPC Observatory Code List</a>"`  to this model.

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+    full_name = models.CharField(max_length=100)
+    short_name = models.CharField(max_length=50)
+    location = models.CharField(max_length=15, choices=LOCATION_OPTIONS)
+    latitude = models.FloatField(
```
jchate6, line 252: Adding help text here (and longitude) explaining that this is a decimal degrees value would be useful.

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+        validators=[MinValueValidator(-90.0), MaxValueValidator(90.0)],
+        null=True,
+        blank=True
+    )
+    longitude = models.FloatField(
+        validators=[MinValueValidator(-180.0), MaxValueValidator(180.0)],
+        null=True,
+        blank=True
+    )
+    elevation = models.FloatField(
```
jchate6, line 262: you need help_text with units for this. (people will use feet...)
also, not saying it will matter, but there are 6 observatories above 5km
https://en.wikipedia.org/wiki/List_of_highest_astronomical_observatories

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+        validators=[MinValueValidator(0.0), MaxValueValidator(50.0)],
+        null=True,
+        blank=True
+    )
+    typical_seeing = models.FloatField(
+        validators=[MinValueValidator(0.0), MaxValueValidator(15.0)],
+        null=True,
+        blank=True
+    )
+    detector_type = models.CharField(max_length=30)
```
jchate6, line 278: Making this an open char field rather than a choice field means that we can't use it to sort by type. 
making this not required, and including a list of basic options will make this much more useful. (Other can be an option.)

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+        null=True,
+        blank=True
+    )
+    orbit = models.CharField(max_length=20, choices=ORBIT_OPTIONS, null=True, blank=True)
+    diameter = models.FloatField(
+        validators=[MinValueValidator(0.0), MaxValueValidator(50.0)],
+        null=True,
+        blank=True
+    )
+    typical_seeing = models.FloatField(
```
jchate6, line 273: help_text units

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+        null=True,
+        blank=True
+    )
+    elevation = models.FloatField(
+        validators=[MinValueValidator(-3000.0), MaxValueValidator(5000.0)],
+        null=True,
+        blank=True
+    )
+    orbit = models.CharField(max_length=20, choices=ORBIT_OPTIONS, null=True, blank=True)
+    diameter = models.FloatField(
```
jchate6, line 268: help_text units

tom_observations/models.py
```diff
@@ -181,3 +183,101 @@ class ObservationTemplate(models.Model):
...
+        validators=[MinValueValidator(0.0), MaxValueValidator(50.0)],
```
jchate6, line 269: I don't feel like we should limit this to 50m.
People may want to include arrays or large radio telescopes.
Arecibo was 300m...

tom_observations/serializers.py
```diff
@@ -58,3 +58,27 @@ def get_queryset(self) -> QuerySet:
             return ObservationRecord.objects.all()
         else:
             return get_objects_for_user(request.user, 'tom_observations.change_observation')
+
+
+class FacilitySerializer(serializers.ModelSerializer):
+    """
+    Class describing the serialization process for a General Facility
+    """
+    site_code = serializers.CharField(max_length=20, required=False)
```
jchate6, line 67: any comments about validation above apply here too...

jchate6 (changes requested):
Looks amazing.
Some comments about adding help text etc.

tom_observations/models.py
rachel3834: I think "description" is too vague for the intended purpose here, which is to be able to search for facilities for specific EM regimes and similar.  
So I've converted this back to a choice field and added help text

jchate6 (approved):
Nice

jchate6 (changes requested):
Need to update limits in serializers

PR #1312, opened 2025-09-23
make simple SOAR forms

jchate6:
image
image

### DIFF
--- a/docs/common/customsettings.rst
+++ b/docs/common/customsettings.rst
@@ -252,6 +252,7 @@ Default:
       'tom_observations.facilities.lco.LCOFacility',
       'tom_observations.facilities.gemini.GEMFacility',
       'tom_observations.facilities.soar.SOARFacility',
+      'tom_observations.facilities.blanco.BLANCOFacility',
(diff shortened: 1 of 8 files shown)


rachel3834 (commented):
Although I like the idea of the separate tabs for the separate instrument modes in general, I found this particular arrangement to be rather confusing.  

The red and blue options are set through a pulldown menu, so why are they split out from spectroscopy generally?  As a user, I am not sure which one I should pick, or why there are multiple options. 

I am strongly in favor of adding default spectrscopic calibrations to the observation groups, but the other thing that puzzles me is why there is a tab for arcs but not for flats?

jchate6:
image
image
@rachel3834 I have renamed the tabs to hopefully be more intuitive. The idea is that there are simple forms specific to the different Goodman configurations. There is nothing in these tabs that cannot be done from the Advanced Spectra tab. I've also added a lamp flat to the auto-generated calibrations.

jchate6:
@griffin-h If you could take a look, would be appreciated.

PR #1320, opened 2025-10-03
Adding blanco facility

### LINKED ISSUE #1307: Add support for Blanco/NEWFIRM to the TOM Toolkit
The LCO portal recently added support for observations on the Blanco Telescope's NEWFIRM spectrograph, which was recently made AEON-compatible.   The first two nights of real-world observations in AEON-mode have been successfully completed.  

The TOM Toolkit should support observations on the Blanco.  This usecase is urgent because of the upcoming Rubin end-to-end alerts demo, at which ideally the goal is to observe with LCO, SOAR, Gemini and Blanco all in AEON mode.  

- [x] For the immediate term, we will make a traditional TOM facility by copying and editing the SOAR facility
- [ ] We are also experimenting with generating facility forms using AEONlib and will see about integrating this into the new BLANCO facility.

jnation3406:
I've tested that the requests validate against the Observation Portal and return a reasonable duration. No tests since this will be throw away code when the new Facilities module is ready.


jchate6 (approved):
Fixed Linting. Merging

PR #1337, opened 2025-10-30
High-eccentricity targets from JPL-H as MPC_COMET

### LINKED ISSUE #1180: JPL Horizons Harvester sets incorrect element scheme for high eccentricity objects
**Describe the bug**
Objects with an eccentricity >~0.9 need to be stored and submitted to the LCO/OCS scheduling system as `MPC_COMET` type due to the need to ultimately use a different mathematical method to calculate positions in the low-level `SLALIB` library. Ingesting comets with high eccentricity such as C/2021 G2 (ATLAS) through the JPL Horizons Harvester  results in an Target created with an orbital element scheme of `MPC_MINOR_PLANET`. This results in a submittal failure when attempting to schedule observations. 

**To Reproduce**
1. Go to 'Targets->Create Targets->Catalog Search'
2. Select 'JPL Horizons' as the Service and enter 'C/2021 G2' in the Term box
3. Click search
4. Observe eccentricity is >1 and scheme is 'MPC Minor Planet'

**Expected behavior**
Objects with an eccentricity >= 0.9  should have scheme set to 'MPC Comet' and the presence of a non-zero/null value for the epoch of perihelion should be verified

**Screenshots**
Attached

Image

- OS: Rocky Linux 9.5
- Browser: Firefox
- Version: 128.7.0esr (64-bit)
- Python Version: 3.11.9

**Additional context**

Fingel:
Makes passing on elements to SLALIB easier when dealing with high eccentricity targets from JPL horizons. Fixes #1180

### DIFF
--- a/tom_targets/forms.py
+++ b/tom_targets/forms.py
@@ -161,7 +161,14 @@ class NonSiderealTargetCreateForm(TargetForm):
         """
         cleaned_data = super().clean()
         scheme = cleaned_data['scheme']  # scheme is a required field, so this should be safe
+        eccentricity = cleaned_data.get('eccentricity')
         required_fields = REQUIRED_NON_SIDEREAL_FIELDS_PER_SCHEME[scheme]
+
+        # Check that eccentiricty isn't too high for a non-comet orbital scheme.
+        if eccentricity and eccentricity > 0.9 and scheme != 'MPC_COMET':
+            raise ValidationError(
+                    "High eccentricity objects should use the MPC_COMET scheme to ensure proper orbital calculations."
+                )
 
         for field in required_fields:
             if not cleaned_data.get(field):
--- a/tom_targets/tests/tests.py
+++ b/tom_targets/tests/tests.py
@@ -428,7 +428,7 @@ class TestTargetCreate(TestCase):
             'epoch_of_elements': 100,
             'lng_asc_node': 100,
             'arg_of_perihelion': 100,
-            'eccentricity': 100,
+            'eccentricity': .1,
             'mean_anomaly': 100,
             'inclination': 100,
             'semimajor_axis': 100,


jchate6 (changes requested):
I think we might want to approach this another way.
I think the logic is the same, but it might make sense to move it into the model. 
I'm picturing moving this to within `save()` or somewhere in `full_clean()` like `model.clean()` so that we perform this on all ingested moving objects.

The problem is that when we try to request an observation for a high eccentricity object we get the wrong pointings if the scheme is incorrect. Doesn't matter where that came from.

jchate6:
I think we have updated the OCS to do this a little better, but this is also an orbital scheme, not a classification or anything. @talister Do you think there is a problem enforcing this change universally? At what level would you expect this to happen?

talister:
> Or maybe this logic really belongs at the LCO/OCS facility level, if what we are really working around is SLALIB? And then we can otherwise leave targets alone.

It's not "working around SLALIB", it's that the maths fundamentally doesn't work in high eccentricity cases and you need to do something different. This is indicated by the scheme which switches (inside SLALIB or whatever else you would use) to a different method of solving Kepler's equation ([background on solving this](https://www.projectpluto.com/kepler.htm) from Bill Gray/`find_orb`). 
Kepler's equation is one of those deceptively simple things but which actually contains a lot of complexity, diverges into chaos (the actual mathematical definition) in certain regimes (high eccentricity) and that excites mathematicians (but see [this](https://www.youtube.com/watch?v=UQHaGhC7C2E)) and is still being worked on ([arXiv paper from 4 months ago](https://arxiv.org/html/2108.03215v3))

jchate6:
@Fingel I've added a validation to the creation form. I think this is good to go, so I'm going to merge it.

PR #1340, opened 2025-10-31
Broker query list and run views should require authentication.

### LINKED ISSUE #855: Anonymous user should not have access to alerts
- [ ] The alerts table should not be visible to anonymous user.
- [ ] Anonymous user should not be able to run alerts.

Related to issue #850

Fingel:
Prevents the bots from accessing the alert list/running alerts. 

There are 2 follow-up improvements that could be made:

1. Make the query run view a POST request. This is more correct. But requires a bit of refactoring as the error views rely on the GET machinery to exist.
2. Make ALL views LoginRequired by default (now the default in Django) and explicitly allow non-authenticated views. This is a much larger change that has the potential to be disruptive if any existing TOMS rely on non-authenticated access to views.

Fixes #855

### DIFF
--- a/tom_alerts/views.py
+++ b/tom_alerts/views.py
@@ -154,7 +154,7 @@ class BrokerQueryFilter(FilterSet):
         fields = ['broker', 'name']
 
 
-class BrokerQueryListView(FilterView):
+class BrokerQueryListView(LoginRequiredMixin, FilterView):
     """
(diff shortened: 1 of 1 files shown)


jchate6 (approved):
Thanks for this. Can you make issues for the remaining improvements you mention?

PR #1346, opened 2025-11-06
remove default slope

jchate6:
Just removes the default. This won't remove the slope for objects made during the previous release.


phycodurus (changes requested):
I've talked to Tim and we've accepted that a default of NULL/None is appropriate here and that FOMO and code around this in the harvesters which is where in FOMO most target creation occurs. So, we could change the default and create a new migration (probably the better way) or modify the migration (to have NULL default values) that created the slope field in the first place.

phycodurus (approved):

PR #1357, opened 2025-11-20
Initial POC for a facility that uses redirects

### LINKED ISSUE #1352: Integrate LCO observing portal into the TOMToolkit workflow
Rather than replicating all of the functionality very cleanly and nicely implemented by the OCS in the observing portal, the TOMToolkit should leverage this implementation.

### iframe
**Pros**
- Fully embedded within TOM
- possible to pass information back to TOM
- Special Obs Portal iframe mode removing top Banners

**Cons**
- Awkward placement (need better control of iframe sizing)
- Double scroll is annoying, Obs portal has arbitrary length

Image

---

### Redirect
**Pros**
- All major changes on the OCS side
- Special redirect Obs Portal mode that pre-loads target/user data, changes top banner, and redirects back to TOM after submit.

**Cons**
- Disruptive. Need to Leave my TOM
- substantial OCS changes needed

Image

---

### New Tab
**Pros**
- Very easy to implement
- Could pass target, user, and post-submit redirect info

**Cons**
- Disruptive, Need to leave my TOM

Fingel:
This is how this flow works:

User is at the TOM:
1. The facility class constructs a URL that points to the OCS. This URL contains query parameters that correspond to the chosen Target's attributes.
2. The facility class also constructs a callback url, which points to the internal tom_observations:callback view. This callback url contains query parameters for the facility class name, the target id, but leaves the observation id blank (this is to be provided by the observation portal)
3. The facility class simply presents a the combined link to the OCS (with the callback url appended as a query parameter). User clicks on the link to the OCS.

Now at the OCS:
1. The OCS presents the observation create form, and pre-fills in the target fields based on the query params.
2. The user submits the observation.
3. Instead of doing whatever the OCS does, it redirects to the callback url created in the previous step 2 with the newly created observation_id appended.

Now back at the TOM:
1. The callback url creates an ObservationRecord object from the query parameters passed in from the OCS: target_id, facility, and observation_id. 
2. The user is sent to the ObservationRecord detail view of the newly created observation record. There is no distinction between records created using the built in OCS form or those created via the redirect flow.


jchate6:
@Fingel is there a branch of the OCS that this relies on to work?

The redirection to the OCS looks good, but I'm guessing this relies on OCS changes to be fully functional. It would be nice to have something running locally before merging this.

tom_observations/facilities/lco_redirect.py
```diff
@@ -0,0 +1,50 @@
...
+    name = "LCORedirect"
+    observation_types = [("Default", "")]
+
+    def target_to_query_params(self, target) -> str:
+        set_fields = {
+            "target_" + k: v for k, v in target.as_dict().items() if v is not None
+        }
+        return urllib.parse.urlencode(set_fields)
+
+    def observation_portal_uri(self) -> str:
```
jchate6: rename to url?

tom_observations/views.py
```diff
@@ -421,6 +434,32 @@ def get(self, request, *args, **kwargs):
         return redirect(reverse('tom_observations:detail', kwargs={'pk': obsr.id}))
 
 
+class ObservationCallbackView(LoginRequiredMixin, View):
```
jchate6: Some Doc strings here would be nice

jchate6 (changes requested):
Looks great. I love the simplicity.

Let's add some small notes for ourselves in the ReadtheDocs about what's expected from a redirect facility on the TOM side and on the Facility side to make this work.

Specifically what the facility should expect and what they need to pass back.

tom_observations/views.py
```diff
@@ -389,6 +390,18 @@ def form_valid(self, form):
         )
 
 
+class ObservationRedirectView(LoginRequiredMixin, View):
+    def get(self, request, *args, **kwargs):
+        facility_name = self.kwargs['facility']
+        facility_instance = get_service_class(facility_name)()
+        target_id = request.GET.get("target_id")
+        callback_url = request.build_absolute_uri(
```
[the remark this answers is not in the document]
Fingel: This is actually Django's method `build_absolute_uri` on a HttpRequest object. They chose it, the mystery lives on... 🤣

PR #1372, opened 2026-01-08
Add a QueryServiceError for dataservices module

Fingel:
This is needed because the underlying query might not be using HTTP directly. Alerce has a client library, so the Alerce Dataservice is using that, which raises it's own custom exceptions.


phycodurus (approved):

jchate6 (approved):
Nice

PR #1373, opened 2026-01-08
Add back crispy field tag for rendering form

Fingel:
Requires us to implement a default for get_layout which excludes the forms related to saving the query.


jchate6 (approved):
Great solution.
Thanks for this!

PR #1401, opened 2026-02-04
Single button for LCO/SOAR/BLANCO configurable text and tooltip

Fingel:
image


Fingel:
> I'm running into a few issues. 1st, I can make the observation request from a non-admin account (which is fine) but I don't have permission to see the observation back in the TOM. This means that the redirect back to the TOM crashes. Ideally I would be able to see observations I created, but at the very least we shouldn't crash.

Reading between the lines here, it sounds like you are using an account on the obs portal that is different than the one the TOM is configured to use. In that case, it makes sense that the redirect would fail because the TOM account doesn't have permission _on the obs portal_ to read that observation. I don't think there is a way around this. TOMs will need to be configured with some kind of super user to be able to access the observation records that all of it's users make on the TOM.
> 
> 2nd, I can't update the status of my observation. When I try to actually update my observation I get `'LCORedirectFacility' object has no attribute 'update_all_observation_statuses'`. This seems to be a method in `BaseRoboticObservationFacility` rather than `BaseObservationFacility` extended by the redirectFacility. This also makes me concerned about `save_data_products` should we actually get some data with this.

Yea, this is an oversight. This facility should delegate to the existing LCO facility for these methods.

jchate6:
> Reading between the lines here, it sounds like you are using an account on the obs portal that is different than the one the TOM is configured to use. In that case, it makes sense that the redirect would fail because the TOM account doesn't have permission on the obs portal to read that observation. I don't think there is a way around this. TOMs will need to be configured with some kind of super user to be able to access the observation records that all of it's users make on the TOM.

This confuses me. I'm talking about the permission to view the observation object created within the TOM Toolkit. Presumably the obs portal passes back the observation request details to the TOM so the observation can be made. Then redirects to the observation detail page in the tom. Why wouldn't the user who initiated the redirect have permission to view the TOM observaiton object that resulted?

jchate6 (approved):
This is working beautifully. 
There is still a lot of work to be done with observations generally, but this is a great way to cover the most general use case.

PR #1402, opened 2026-02-04
Default to LCORedirect facility for new TOMS

Fingel:
This disables the LCO/SOAR/BLANCO facilities by default in new TOMs, and enables the RedirectFacility while keeping compatibility with existing TOMs that may already have observations using the current facilties.

Disabling the current facilities would cause crashes as observations are linked to the facility they are created from.


jchate6 (approved):
We need to wait for 1401 to actually publish, but looks good in that context.

jchate6 (changes requested):
I'll go ahead and fix this.

jchate6 (approved):

PR #1407, opened 2026-02-11
Add new Scout interface

### LINKED ISSUE #1431: Migrate Scout Broker to data service

talister:
This (draft) PR creates an interface to the JPL Scout service via its API endpoint. This is implemented using the new(ish) `tom_dataservices`. It is substantially functional in terms of fetching, filtering and creating targets but some questions before it's mergeable:

1. Having always struggled with Mocking and monkeypatching of `requests`-based functionality, it's not clear how best to test this without hitting live services
2. I've not seen a way to get access to the `request` to add a message to through the Django `messages`  infrastructure, if e.g. the user searches for a target that is removed 
3. The `Target`s can be created in the initial run of fetching targets but will need to be updated with new values for existing `Target`s if Scout reruns on the same data producing different results or the number of observations changes. I think generally `Target`s in the TOM Toolkit once created are relatively static and only checking on the `name` is done to avoid creating duplicates. I could sub-class the `match_target()` to match on things other than `name` but the NEOx experience has taught me that checking for changes in e.g. the floating point orbital elements can be tricky. Any thoughts on how to do robust updating of changing Targets?


tom_dataservices/data_services/jpl.py
```diff
@@ -0,0 +1,260 @@
+from math import sqrt, degrees
+
+from astropy.constants import GM_sun, au
+from django import forms
+from django.contrib import messages
```
jchate6: Lint: Remove this import

jchate6 (changes requested):
Lint fixes.

jchate6:
All of the scout changes have been pulled and put into https://github.com/TOMToolkit/tom_jpl

All that remains here is some basic improvements to dataservices.

tom_dataservices/templates/tom_dataservices/query_form.html
```diff
@@ -7,8 +7,8 @@ <h2><a href="{{info_url}}" target="_blank">{{ form.data_service.value }}</a> Que
 {% else %}
 <h2>{{ form.data_service.value }} Query Form</h2>
 {% endif %}
-{% if verbose_name != form.data_service.value%}
-    <h4>{{ verbose_name }}</h4>
+{% if get_verbose_name != form.data_service.value %}
```
phycodurus: Is the intention here really about the what name to display (is whether it's verbose or not really the issue?) (i.e the "display_name")? Could a `@property` like `display_name` on the `DataService` class contain this logic? That way, the for form.html just displays the display_name. (also, I think the convention for property naming doesn't involve `get_` So, the "call site" looks like `instance.computed_property`

tom_dataservices/dataservices.py
```diff
@@ -231,11 +242,11 @@ def query_spectroscopy(self, query_parameters, **kwargs):
 
     def query_reduced_data(self, target, **kwargs):
         """Set up and run a specialized query to retrieve Reduced Datums from a Data Service"""
-        build_query_parameters_from_target_method = getattr(self, 'build_query_parameters_from_target')
+        build_query_parameters_from_target_method = getattr(self, 'build_query_parameters_from_target', None)
```
phycodurus, line 239: Not a big deal, but given how the default (`None`, atm) is being used, defaulting to `False` might be more direct/type-safe.

tom_dataservices/dataservices.py
```diff
@@ -309,7 +320,8 @@ def to_reduced_datums(self, target, data_results=None, **kwargs):
             raise MissingDataException('No Reduced Data dictionary found.')
         reduced_datum_list = []
         for key in data_results.keys():
-            reduced_datum_list += self.create_reduced_datums_from_query(target, data_results[key], key, **kwargs)
+            if data_results[key]:
```
phycodurus, line 318: I think this logic needs a comment providing context. Something like "# a Falsey value for data_results[key] means <what does this mean?>" or "# only create reduced_dataums if there are reduced_dataums to create" or ??. Just don't make the reader have to guess what this means.

phycodurus (approved):

tom_dataservices/templates/tom_dataservices/query_form.html
jchate6: This was poorly done.
I have fixed it to do what I actually want.
I really want a "name" / "Verbose name" and to have that display behavior make sense.

Image

The purpose of get_verbose_name was to allow us to return a verbose name all the time, but have it just default to name if it didn't exist. But I don't think that's worth it here.

PR #1414, opened 2026-02-18
Bugfix/fix single obs cadence strategy

### LINKED ISSUE #1413: Tom observations cadence RetryFailedObservationsStrategy isn't updating dynamic cadence
**Describe the bug**
The single observation strategy RetryFailedObservationsStrategy in `tom_observations` does not update the `DynamicCadence` object associated with the observation group to set `active = False` upon completion. This causes the cadence to show up and clutter `./manage.py runcadencestrategies` even though the observation is not active anymore upon completion. The strategy also does not update the observation group in the database to reflect the observation status from OCS. When an observation is submitted and `./manage.py runcadencestrategies` is run, the observation does not reflect a `'PENDING'` observation states. The repeating strategy (`ResumeCadenceAfterFailureStrategy`) does not reflect this bug and does update the status to `PENDING` and since it is only stopped on cancellation, it does not experience the issue with the `DynamicCadence` needing to be turned off upon a completed observation.

**To Reproduce**

1. Go to a target page on TOM Demo or a fresh TOM that has the form
2. Under "Observe" click on "LCO" then open the "Photometric Sequence" form, or select an observation template
3. Select "Once in the next" or "RetryFailedObservationsStrategy" under "cadence strategy"
4. Input observing parameters and submit
5. Upon completion of the observation, the sequence will still appear in the Dynamic Cadence table and when running `./manage.py runcadencestrategies`

**Expected behavior**
Upon observation request (or when `runcadencestrategies` is run), the `ObservationRecord` status should update from `''` to `'PENDING'`. Similarly updating to `'COMPLETE'` upon completion. When the observation is complete, the `DynamicCadence` should be turned off. If a failure in the observation occurs, the observation should be re-requested with a shifted window and is only finished when a complete state is reached.

moira-andrews:
See #1413, I've tested this in the SNEx2 environment with success. I didn't change the logic of handling failed observations, but added the check used in `ResumeCadenceAfterFailureStrategy` to update the observation status along with a check if that observation is completed, to turn off the dynamic cadence.


jchate6 (changes requested):
first things first, this is now hitting several more live services via `form.is_valid()` on line 57. 
Any chance you can take a crack at mocking these in the test, or limiting these some other way?

Also, additional tests for what happens for these various conditions would be appreciated, and help document the code.

moira-andrews:
Addressed in #1592

PR #1423, opened 2026-02-24
runbrokerquery management command should not run forever

griffin-h:
This changes the `manage.py runbrokerquery` command from a `while` loop to a `for` loop, so that it does not continue running forever without re-querying the broker. It should likely be run on a cronjob.


jchate6 (changes requested):
Can you remove the unused `sleep` import?

PR #1425, opened 2026-02-25
Update Django 5.2.11

### LINKED ISSUE #1349: Investigate supporting Django 5 and 6
Django 5 has been out for a while. We should consider supporting it if possible.

Fingel:
The edit to session_utils is because Django 5 no longer supports using unsaved objects in query expressions.

The edits to tom_targets/tetss/tests.py are simply a renamed function.

Notable features of Django 5 relevant to the TOM Toolkit:

* [Database computed defaults](https://docs.djangoproject.com/en/6.0/releases/5.0/#database-computed-default-values) allows generating defaults values from the database, not python. This might come in handy in specific database-heavy areas, like PostGIS or Healpix.
* The new [ClosestPoint()](https://docs.djangoproject.com/en/6.0/releases/5.0/#django-contrib-gis) function sounds like it could be helpful in RA/DEC searches, perhaps for non-exact/fuzzy search where only a single result is desired.
* Some improvements to [pagination errors](https://docs.djangoproject.com/en/6.0/releases/5.0/#pagination) which might make the HTMX table stuff smoother.


phycodurus (approved):

jchate6 (approved):
nice

PR #1426, opened 2026-02-25
1396 update reduced datums for existing target via data service

jchate6:
Adds Redundancy for new data from Data Services
Adds Docs
Adds form to Target detail page to Target Detail page to get new data from Data Service.

PR #1427, opened 2026-02-26
Dataservices alerce

### LINKED ISSUE #1367: Migrate Alerce Broker to DataServices arch
for details see 
- https://github.com/TOMToolkit/tom_base/issues/1288
- #1376

Fingel:
Includes branch #1426

Implementation of the Alerce data service, including creating reduced datums.


jchate6 (changes requested):
This isn't a full review, but I wanted to start the discussion about dependencies.

This requires that we add the alerce package to our list of dependencies. 
Do we have any qualms about that? Should this be pulled into its own installable app?

Fingel:
> This isn't a full review, but I wanted to start the discussion about dependencies.
> 
> This requires that we add the alerce package to our list of dependencies. Do we have any qualms about that? Should this be pulled into its own installable app?

If there is a precedent for putting functionality that requires a 3rd party module into an installable app, then yes. But if we go off of the "weight" of the dependency the alerce client only depends on pandas, astropy and requests, so those are likely already installed.

jchate6 (changes requested):
@Fingel I still haven't done a full code review. I have merged in the most recent dataservice updates and made a few required changes. Just going through the workflow I've hit a few things that need work. I'm going to start working on some of these, so I'll check things off as I go, and be sure to pull and check in if you want to work on some of this too.

Changes that need to be implemented:

- [x] Fix form layout. We include a layout, but this is overridden by the advanced form partial. If we want to include instructions and links, we need to either not include the partial or add the layout features to the partial.
- [x] The Broker let you filter by location, time, number of detections, etc as well as set the number of objects to retrieve and how they should be sorted. Is there some way to get this functionality into the Dataservice? Does the ALeRCE client not allow for any of these features
- [x] Survey doesn't seem to affect classifier queries. This should be made obvious if on purpose or fixed if not.
- [x] Organize advanced form to be a little more readable.
  - [x] break into classifier sections, with each one containing a choice field and probability with classifier title and help text.
  - [x] Add some range for the probability, (is this a 0-1 or a 0-100? what's default?)
  - [x] what does the help text mean? (hierarchical_random_forest_1.0.0, stamp_classifier_1.0.4) is this the API version that is being referenced? Is that useful information for the user?
- [x] Clean up the results table
  - [x] add custom results table
  - [x] convert column headers to be human readable
  - [x] leave off less useful data.
- [x] Include lc data on target creation
- [x] Fix lc query
- [ ] Fix Forced Phot query. (Should be called with more intention since it's expensive.)
- [x] Reduced Datums get_or_create not behaving as expected.

jchate6 (approved):

PR #1437, opened 2026-03-01
Calendar

### LINKED ISSUE #1436: Basic calendar functionality
Before implementing TOM/Astronomy specific functionality to the calendar, a baseline version with the stuff someone would expect from a calendar should be built out.

* Month view
* CRUD operations on calendar events
* Display types for single and multi-day events

I added the moon phase to each day because it could be helpful for some programs and we already have all the tools in the TOM to calculate it simply.

https://github.com/user-attachments/assets/d8d66e4f-0dd8-4cbb-bae8-206caac36585

### LINKED ISSUE #1442: Associate target lists with calendar events
There should be an association with target list (group?) with a calendar event. The calendar view might show a count of targets with the associated list, with the full list displayed in the calendar update modal.

### LINKED ISSUE #1443: Calendar - more fields
Add some free-form fields for calendar events:
proposal, user and telescope for now.

### LINKED ISSUE #1444: Calendar - todo list
Add todos that can be marked as completed and associated with a calendar event. The calendar display should show the count, maybe, if there's space.

Fingel:
Basic calendar application as a base to build TOM/Astronomical specific functionality on top of. Just a simple month view for now, with the standard CRUD actions for events. Event model can be extended. Event model can be created elsewhere in the TOM as an easy way to add items to the calendar.

Uses HTMX, no third party dependencies. 

Closes #1436 #1442 #1443 #1444 

https://github.com/user-attachments/assets/526f1921-2a31-4b07-80bb-25669d104fb2


rachel3834:
Bug:
-  [x] Clicking on an event with an URL redirects to that URL; this should bring up the edit popup
-  [x] Clicking on a random day with no event brings up the last-created event or nearest-event edit box.

Feedback on initial implementation of calendar from TOM Hack Session:
- [ ] When creating an event entry in the calendar, it would be great to have another string field for the instrument as well as the telescope. 
- [ ] In the event create popup box, datetime fields should indicate the timezone they refer to.  Having UTC as the default is ideal, but most users would also consider setting this to the telecsope's local time (though UTC is the only thing that works for space telescopes).  Also, the field should indicate the expected format for each of the components of the fields. 
- [ ] It would be great to be able to add "to dos" when creating the event and not just when editing an existing one
- [ ] It would be great to be able to reorder to the to-dos, or edit the list to insert ones.
- [ ] It would be great to support different types of events with different parameters.  The current one is great for observing runs, but users would also like to add, e.g.:
- [ ]  Meetings with location fields rather than telescope and instrument
- [ ] Deadline 
- [ ] A direct link to the Target List rather than just the name of the list would be helpful. (We did notice the little vertical bar on the calendar block but we think this is a bit subtle)
- [ ] Is it possible to have day and week displays for the calendar as well as monthly?  Alternatively clicking on a day to bring up an overview of all events on that day would be useful. 
- [ ] Is it possible to query the calendar via API or to add events to the calendar?

For future calendar in target page:
- Days to indicate angular separation of target from Moon.

jchate6 (changes requested):
When clicking on an empty day, it should bring up the new event pop up rather than the most recently selected event.

This appears to be caused by a value error:
`ValueError: time data 'April 16, 2026' does not match format '%Y-%m-%d'`

There are also a bunch of codacy concerns.

Fingel:
> When clicking on an empty day, it should bring up the new event pop up rather than the most recently selected event.
> 
> This appears to be caused by a value error: `ValueError: time data 'April 16, 2026' does not match format '%Y-%m-%d'`

I can't reproduce this in Firefox or Chromium. Create new event, close. Click blank area on calendar and it's a fresh form. Same for viewing another event. Is there a specific sequence I need to follow to trigger this? 

> 
> There are also a bunch of codacy concerns.

These are awful!  "Missing docstring for magic method __str__"?  Really? It also wants a docstring for the `Meta` class on the Event model... just no.

There's a few others, all related to docstrings, about  various missing newlines or whatever. I tried fixing a few but this seems like mostly a waste of time.

jchate6:
> I can't reproduce this in Firefox or Chromium. Create new event, close. Click blank area on calendar and it's a fresh form. Same for viewing another event. Is there a specific sequence I need to follow to trigger this?

Yeah, I'm not sure what's going on, I didn't look too closely into it. 
@rachel3834 seemed to mention a similar issue above. I'll take a more thorough look next week.

> These are awful! "Missing docstring for magic method **str**"? Really? It also wants a docstring for the `Meta` class on the Event model... just no.
> 
> There's a few others, all related to docstrings, about various missing newlines or whatever. I tried fixing a few but this seems like mostly a waste of time.

I think we must have added the  pydocstyle linter or something.
I didn't really look at the issues when I made that comment. Since then, I've taken a closer look and we should figure out how to disable this. It triggered on 2 contradictory style formats (see slack) so something is definitely not configured correctly.

jchate6 (approved):
I've made a small change to keep the app from being mandatory.

PR #1446, opened 2026-03-11
Match path prefixes in locked auth strategy

### LINKED ISSUE #1445: Adding password reset to login page
For SNEx2, I have tried to add a password reset link on the login page. I have had success accessing the Django default password reset url by adding a template `/registration/login.html` that includes `<a href="{% url 'password_reset' %}" class="text-muted">Forgot your password?</a>` and `path('accounts/', include('django.contrib.auth.urls')),` to my `urls.py`. I've also updated the `settings.py` to include the password_reset urls. 

```
AUTH_STRATEGY = 'LOCKED'
OPEN_URLS = [
    '/accounts/password_reset/',
    '/accounts/password_reset/done/',
    '/accounts/reset/done/',
    '/accounts/reset/'
]
```

The issue I've run into with this is I am able to follow the url to reset my password and input my email. From my email I will get a link with `/accounts/reset/<token link>` and when I paste this into my url, I get an error that I do not have permission to access the page. When I change `AUTH_STRATEGY=READ_ONLY`, I am able to access the url `/accounts/reset/<token link>` sent in the email and change my password. I've tried variations of '/accounts/reset/' in open urls while having `AUTH_STRATEGY = 'LOCKED'` like '/accounts/reset/*' or even just only including `/accounts/`, but the only successful method was changing the `AUTH_STRATEGY`.

Fingel:
Resolves #1445 by ensuring that urls like /accounts/reset/<id> can be matched in open_urls with /accounts/reset/

### DIFF
--- a/docs/common/customsettings.rst
+++ b/docs/common/customsettings.rst
@@ -165,8 +165,8 @@ details and available hooks.
 Default: []
 
 With an `AUTH_STRATEGY <#auth-strategy>`__ value of **LOCKED**, urls in
-this list will remain visible to unauthenticated users. You might add
-the homepage (‘/’), for example.
+this list will remain visible to unauthenticated users. You can also use wild cards to open an entire path.
+You might add the homepage (‘/’), for example, or anything with a path that looks like '/accounts/reset/*/'.
 
 `TARGET_PERMISSIONS_ONLY <#target-permissions-only>`__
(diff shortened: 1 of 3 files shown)


jchate6 (changes requested):
We will also want to update the docs in `common/customsettings.rst` and the comment in `tom_setup/templates/tom_setup/settings.tmpl:331,351` and `tom_base/settings.py`

jchate6 (approved):

PR #1466, opened 2026-03-20
Lazy load facility status page and use htmx for refresh

davner:
This uses lazy loading for the telescope facility page and adds a button to refresh the facility status. A django message is displayed only when using the refresh button. 

This makes TOMToolkit more reactive.


tom_observations/templates/tom_observations/facility_status.html
```diff
@@ -5,6 +5,22 @@
...
+  <div
+    hx-get="{% url 'tom_observations:render-facility-status-list'%}?hx_trigger=load" hx-target="#facility-status-list"
+    hx-indicator=".progress"
+    hx-trigger="load"
+    hx-disabled-elt="this">
+    <button
+      class="btn btn-secondary mb-3"
+      hx-get="{% url 'tom_observations:render-facility-status-list' %}"
+      hx-trigger="click"
+      hx-disabled-elt="this">
```
Fingel, line 16: wha wha wha what is this!???

PR #1479, opened 2026-03-26
Convert the MPC Explorer Harvester to a DataService

### LINKED ISSUE #1433: Migrate MPC Harvester to data service

talister:
Implements #1433


tom_dataservices/templates/tom_dataservices/mpc/partials/mpc_results_table.html
```diff
@@ -0,0 +1,36 @@
...
+        <th>Argument of Perihelion (ω)</th>
+        <th>Epoch of Perihelion (t_p)</th>
+        <th>Absolute Magnitude (H)</th>
+        <th>Slope Parameter (G)</th>
+    </tr>
+    </thead>
+    <tbody>
+    {% for result in results %}
+    <tr>
+        <td><input type="checkbox" name="selected_results" value="{{ result.id }}"/></td>
```
jchate6, line 21: This table is quite wide. How do we feel about a visual truncation of these values:

Image

The actual ingested values are full length.

tom_dataservices/data_services/mpc.py
```diff
@@ -0,0 +1,194 @@
...
+        if result['designation_data'].get('name', "") != "":
+            extra_desigs.append(result['designation_data']['name'])
+        extra_desigs.append(unpacked_primary_desig)
+        extra_desigs += result['designation_data']['unpacked_secondary_provisional_designations']
+        # Make sure we don't include the primary designation twice
+        try:
+            extra_desigs.remove(target.name)
+        except ValueError:
+            pass
+        target.extra_names = extra_desigs
```
jchate6: This isn't the best way to do this.  We will probably want a `create_aliases_from_query`.  I'll work on a solution to get additional names in tomorrow.

jchate6 (commented):
I'm working on cleaning up a few things, but did have a question.

tom_dataservices/templates/tom_dataservices/mpc/partials/mpc_results_table.html
talister: If It Is Even Possible (Tm), I think I would like to do a similar thing as I did with the FOMO ephemeris generation where I had a 'high precision ?' toggle button of the form (which can default to off) and round or not based on that setting. I think 5 d.p. for the elements and 2 d.p. for _H, G_ should be the default for rounded.

tom_dataservices/data_services/mpc.py
talister: Agreed, the multiple names minor bodies can have is complex enough to warrant a separate (testable) method

PR #1492, opened 2026-04-01
Update to the ResumeCadenceAfterFailureStrategy

moira-andrews:
Updated the observation payload to have `observation_payload['scheduled_start'] = last_obs.scheduled_start and observation_payload['scheduled_end'] = last_obs.scheduled_end` with `'scheduled_start'` and `'scheduled_end'` as the start and end keywords to be passed into `advance_window`.


moira-andrews:
Addressed in #1592

PR #1498, opened 2026-04-02
shift start of cadence to observation time

### LINKED ISSUE #1497: Discrepancy between the start time for a cadence and the time of actual observation
Currently, the ResumeCadenceAfterFailureStrategy triggers a new cadence based on the start of the initial window rather than the actual successful observation. 
Users would like to change this to more accurately reflect the preferred behavior of the next cadence beginning at the end of the observation.

i.e. if I want an observation every 7 days, but give myself a 3 day window, I want my next window to start 7 days after the successful observation, not after the beginning of my window.

jchate6:


tom_observations/cadences/resume_cadence_after_failure.py
```diff
@@ -105,7 +107,14 @@ def advance_window(self, observation_payload, start_keyword='start', end_keyword
...
 
-        new_start = parse(observation_payload[start_keyword]) + timedelta(hours=advance_window_hours)
+        # If the last observation has an end or start time recorded use that for the beginning of the new cadence
+        # Otherwise use the start time of the current cadence.
+        if observation_payload['last_obs_end']:
+            new_start = observation_payload['last_obs_end'] + timedelta(hours=advance_window_hours)
+        elif observation_payload['last_obs_start']:
+            new_start = observation_payload['last_obs_start'] + timedelta(hours=advance_window_hours)
+        else:
+            new_start = parse(observation_payload[start_keyword]) + timedelta(hours=advance_window_hours)
```
moira-andrews, line 117: I worry about the window length being limited to the length of the observation because if it is a short exposure that isn't constantly visible, it will clutter up the `runcadencestrategies` log with visibility errors as it keeps attempting to schedule it. This also might be tricky with many observations because it's more observations being scheduled for the same window instead of giving the scheduler room to avoid failures. Maybe a combination of this more robust check for scheduled start/end method and a user defined from the settings file for the minimum window

moira-andrews: This is definitely the right direction though, and cluttering runcadencestrategies log is not the worst thing in the world. Maybe a future improvement would be to combine with the method in #1492 to allow the observing window to be increased or defined elsewhere

PR #1506, opened 2026-04-09
bump specutils minimum required version to >2.0

miguelgomezgomez:
This update is required for integration with GOATS and the dependencies of its other modules


jchate6 (approved):
Thanks for this PR!

I've changed the version dependencies a bit to be more in line with our practices.

PR #1508, opened 2026-04-10
add default_settings so users don't have to update their settings when new mandatory apps or middleware is used.

### LINKED ISSUE #1507: We need a way to manage required apps
Tomtoolkit should be able to add a new required app (internal or 3rd party) without users having to update their settings.py.

It would make sense for us to have some form of "default settings" file that can be loaded into a users settings to include all of the mandatory changes that are needed in a new TOM version transparently to the users.

jchate6:
This should be added to the list of things users need to update for v3


tom_setup/templates/tom_setup/settings.tmpl
```diff
@@ -38,51 +41,13 @@ TOM_NAME = '{{ PROJECT_NAME }}'
...
-    'django_filters',
-    'django_tables2',
-    'django_gravatar',
-    'django_htmx',
-    'tom_targets',
-    'tom_alerts',
-    'tom_catalogs',
-    'tom_observations',
-    'tom_dataproducts',
+INSTALLED_APPS += [
```
phycodurus: I like the idea of:
```
INSTALLED_APPS = TOMTOOKIT_INSTALLED_APPS + [your-toms-installed-apps]
```
and leave the naming up to you :-)

tom_setup/templates/tom_setup/settings.tmpl
```diff
@@ -38,51 +41,13 @@ TOM_NAME = '{{ PROJECT_NAME }}'
...
-    'django.middleware.common.CommonMiddleware',
-    'django.middleware.csrf.CsrfViewMiddleware',
-    'django.contrib.auth.middleware.AuthenticationMiddleware',
-    'django.contrib.messages.middleware.MessageMiddleware',
-    'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'django_htmx.middleware.HtmxMiddleware',
-    'tom_common.middleware.Raise403Middleware',
-    'tom_common.middleware.ExternalServiceMiddleware',
-    'tom_common.middleware.AuthStrategyMiddleware',
+MIDDLEWARE += [
```
phycodurus: same idea:
```
MIDDLEWARE = TOMTOOLKIT_DEFAULT_MIDDLEWARE + [any-additional-middleware-you-TOM-needs]
```
again, naming up to you (but it should probably be consistent like my examples are not :-).

phycodurus (approved):

PR #1512, opened 2026-04-16
Fix hermes ingestion so it ingests targets and spectroscopic data too.

jnation3406:
The hermes alertstream ingestion was just a barebones implementation before, only ingesting photometry if a target matched by name. I've updated it to match targets by name or by alias name, or create a new target with that name in the TOM. I've also added the ability to ingest spectroscopic data in both ways that it can be sent through HERMES - as a linked file, or as raw data arrays.

I had to separate the `alertstreams.hermes.py` class into a `hermes_publisher.py` and `hermes_ingester.py` to get around a circular import issue - sharing uses hermes publishing, but ingesting uses data_processing which uses sharing.

I've tested all this locally with hermes-dev and its doing what it is supposed to. It involves DataProducts, ReducedDatums and data processors, so I'm sure lots will need to be updated when @Fingel 's overhaul is done.


tom_dataproducts/templatetags/dataproduct_extras.py
```diff
@@ -383,18 +383,19 @@ def spectroscopy_for_target(context, target, dataproduct=None):
         spectroscopy_data_type = settings.DATA_PRODUCT_TYPES['spectroscopy'][0]
     except (AttributeError, KeyError):
         spectroscopy_data_type = 'spectroscopy'
-    spectral_dataproducts = DataProduct.objects.filter(target=target,
```
jnation3406, line 386: This change allows us to support showing plots of spectroscopy ReducedDatums without an accompanying DataProduct, the same way we do with photometry datums.

PR #1521, opened 2026-04-23
Reducedddatum refactor

### LINKED ISSUE #609: Make Data validation and control class for Reduced Datums

### LINKED ISSUE #1505: Refactor ReducedDatums to a multi-model based infrastructure
We want to refactor Reduced Datums to have more structure.
We will have 4 types of Reduced Datums in tom Base:

- General (Anything not otherwise covered)
- Photometry
- Spectroscopy
- Astrometry

Users are expected to create their own models for more complex forms of calculated data.

See:
 - #609 
 - #1480 
 - [Model Field Breakdown Sheet](https://docs.google.com/document/d/14RFJHyewhR4qh61H5WLVUVu6NlRj6nhSJIDRtorMeBE/edit?usp=sharing)

Fingel:
Introduces concrete Reduced Datum types: PhotometryReducedDatum, SpectroscopyReducedDatum, AstrometryReducedDatum, and keeps the original ReducedDatum class as a generic type. This PR attempts to use the concrete types where needed, while maintaining API compatibility. This means the the JSON endpoints will have the same shape, but the underlying storage will use the concrete tables.

It also includes a management command to move existing generic ReducedDatums into their concrete counterparts. Note that many views will no longer populate until this command is run, for example the photometry plot.


tom_dataproducts/models.py
```diff
@@ -311,137 +313,286 @@ def create_thumbnail(self, width=None, height=None):
...
     :param message: Set of ``AlertStreamMessage`` objects this object is associated with.
     :type message: ManyRelatedManager object
 
     """
 
     target = models.ForeignKey(BaseTarget, null=False, on_delete=models.CASCADE)
-    data_product = models.ForeignKey(DataProduct, null=True, blank=True, on_delete=models.CASCADE)
-    data_type = models.CharField(
-        max_length=100,
-        default=''
```
jchate6, line 394: Should we keep this validate_unique for the generic ReducedDatum? It seems like a reasonable safe guard

tom_dataproducts/models.py
```diff
@@ -311,137 +313,286 @@ def create_thumbnail(self, width=None, height=None):
...
+    :type source_location: str
+    """
+
+    data_type = models.CharField(max_length=100, default="")
 
     class Meta:
-        get_latest_by = ('timestamp',)
+        get_latest_by = ("timestamp",)
 
     def save(self, *args, **kwargs):
```
jchate6: I think we should get rid of this validation error. There is no reason to restrict the data_type to the DataProduct types anymore. Part of the reason for that was because we didn't have a way to tell if something was phot or not without that. Now we do.

tom_alerts/brokers/alerce.py
jchate6, line 1: This code will be abandoned in the next sprint

tom_alerts/brokers/gaia.py
jchate6, line 1: This will probably be useful when we refactor this into a data service. Thanks!

tom_dataproducts/management/commands/migrateReducedDatums.py
```diff
@@ -0,0 +1,140 @@
...
+        ra_error_units=value.get('ra_error_units', ''),
+        dec_error_units=value.get('dec_error_units', ''),
+    )
+
+
+class Command(BaseCommand):
+    help = (
+        'Migrates generic ReducedDatum rows into their concrete typed models '
+        '(PhotometryReducedDatum, SpectroscopyReducedDatum, AstrometryReducedDatum) '
+        'and deletes the originals. Run this once after deploying the reduceddatum refactor.'
```
jchate6: Let's add a comment about this being available in v3

jchate6 (changes requested):
I've made a few preliminary comments.

I also merged v3 into this. Can you take a look at tom_dataservices to see what changes might need to happen there?

Once you are happy with that, please re-request a review.

tom_dataproducts/models.py
```diff
@@ -335,113 +337,288 @@ def get_queryset(self):
...
 
-    :param data_type: The type of data this datum represents. Default choices are the default values found in
-        DATA_PRODUCT_TYPES in settings.py.
-    :type data_type: str
-
-    :param source_name: The original source of this datum. The current major use of this field is to track the broker a
-                        datum came from, but can be used for other sources.
-    :type source_name: str
-
-    :param source_location: A reference to the location that this datum was originally sourced from. The current major
```
jchate6, line 358: We probably want to keep this docstring since the field still exists.

tom_dataproducts/models.py
Fingel: I've added it back, but remember this is the code that does not scale (need to introspect the value field of every row in the database on every save). But as long as we only expect to scale on the concrete types I guess it's fine.

jchate6 (approved):
I think this is ready.

PR #1533, opened 2026-04-30
Skip re-encrypt when _password is unavailable

Fingel:
Fixes an issue when trying to login. We are using _password which doesn't appear to be guaranteed to be set by Django. This seems to happen after session expiration.


jchate6 (approved):
Good stop gap.

PR #1534, opened 2026-04-30
Show indicator on sortable columns with no sort applied

### LINKED ISSUE #1411: HTMX Table visuals need some improvement
It should be more obvious how to use the sorting functionality.
- [ ] We need to show which column the table is sorted by from the outset
- [ ] We need to have faded stand ins for the sortable icons to let people know the table is sortable. (up/Down arrows)

Fingel:
Shows the double arrows on columns which are sortable but don't have sorting applied.
From the issue:

> We need to show which column the table is sorted by from the outset

This works already, the issue is the target table is sorted by "-created" by default and that's not actually a column in the table. I tested that it does work by changing the default sort to "-name" and the column appears with the down arrow as expected.

image

Fixes #1411

### DIFF
--- a/tom_common/static/tom_common/css/main.css
+++ b/tom_common/static/tom_common/css/main.css
@@ -131,14 +131,23 @@ td, th {
 }
 
 /* add up and down sorting arrows to column headers  */
-.table-container th.asc::after {
-    content: '\0000a0\0025bc';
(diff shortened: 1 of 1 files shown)


jchate6 (approved):
Nice. Now we just need it smart enough to know that M110 doesn't go between M11 and M12...

PR #1545, opened 2026-05-06
add range for ndet

### LINKED ISSUE #1543: Final Alerce Touches
Just a few final things to address for ALeRCE Data Service

- [x] add a max number of detections field

jchate6:
ndet takes a range: https://alerce.readthedocs.io/en/latest/apis/alerce.html#alerce.ztf_search.ZTFSearch.query_objects

So this will allow for the number of detections to be within a max/min

### DIFF
--- a/tom_dataservices/data_services/alerce.py
+++ b/tom_dataservices/data_services/alerce.py
@@ -32,7 +32,8 @@ class AlerceForm(BaseQueryForm):
     firstmjd_lt = forms.FloatField(required=False, label="Max MJD of first detection")
     lastmjd_gt = forms.FloatField(required=False, label="Min MJD of last detection")
     lastmjd_lt = forms.FloatField(required=False, label="Max MJD of last detection")
-    ndet = forms.IntegerField(required=False, label="Min. Number of Detections")
+    ndet_min = forms.IntegerField(required=False, label="Min. Number of Detections")
+    ndet_max = forms.IntegerField(required=False, label="Max Number of Detections")
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


tom_dataservices/data_services/alerce.py
```diff
@@ -149,6 +155,7 @@ def query_service(self, query_parameters, **kwargs) -> list[dict]:
             params["dec"] = dec
             params["radius"] = radius
         results = []
+        print(params)
```
phycodurus: should probably be a `logger.debug` call

tom_dataservices/data_services/alerce.py
```diff
@@ -136,8 +137,13 @@ def query_service(self, query_parameters, **kwargs) -> list[dict]:
             lastmjd_lt := query_parameters.get("lastmjd_lt")
         ):
             params["lastmjd"] = [lastmjd_gt, lastmjd_lt]
-        if ndet := query_parameters.get("ndet"):
-            params["ndet"] = ndet
+        ndet_min = query_parameters.get("ndet_min")
```
phycodurus: this logic could probably be made more obvious or at least add a comment that says what's happening here. (My preference is to have comments that say what the code is doing so I can understand what's going on by reading the comments. If I want to know _how_ what going on is being done, then I'll read the code). 

Also, `assignment expression` (`:=`) will work with the `elif` (like `if`), if you want to use it here.

phycodurus (approved):
All my comments are about comments and code style. (so, fwiw). Except there is a `print` statement in there.

Fingel (approved):

tom_dataservices/data_services/alerce.py
phycodurus, line 126: I think it would be good in the docstring to provide an example of what the `params` dictionary actually looks like. (unless you want to introduce an Pydantic BaseModel subclass here). 

the value of the 'ndet` key in particular is confusing because sometimes it's a list of length 1 and sometimes it's a list of length two. Whoever looks at this code in the future isn't going to know what's up with that without some research.

PR #1546, opened 2026-05-06
Assign correct target permissions when using a custom target model

### LINKED ISSUE #1538: Error when trying to assign target to group as non-admin user
**Describe the bug**
When a non-admin user (in our case: a user that has registered through tom-registration and has been approved by an admin) creates a target group and tries to add a target  to that group, the following error message appears:

Image

**To Reproduce**
1. Login as non-admin user
2. Create a target group (go to /targets/targetgrouping/, click "create new grouping", fill out the form, etc.)
3. Go to /targets
4. Select n target(s), make sure target group is selected and click "add"
5. You see a success message that says 0 targets were added and a message about permissions and app_label mismatchs

**Expected behavior**
- if the user does not have the permissions to add targets to a group, a message saying "permission denied" should appear
- if the user has the permission, the targets should be added to the group and a success message should appear.

**Screenshots**
Image

- OS: MacOS
- Browser: Firefox
- Version [e.g. 22]: Firefox 150.0
- Python Version: Python 3.11.15

**Additional context**
- we are using a custom target model that extends BaseTarget
- I have looked at the code that produces the error message; seems to be here https://github.com/django-guardian/django-guardian/blob/main/guardian/backends.py#L107.
Guardian compares app_labels and these do not match, so this seems to be the cause of the error. I don't know, if we can just "change" the app_label in our Meta class in our custom model to be the same as BaseTarget or if that has some negative consequences. As a fallback, guardian compares the app_label of the content type and allows providing a custom function for fetching the content type via `guardian_settings.GET_CONTENT_TYPE`. I did try this with the following approach and it got rid off the error message and now just displays "Failed to add target with id=foo to group 'foo'; Permission denied. "
```python
def get_content_type(obj):
    if type(obj) is GalacticTarget:
        base_target_instance = BaseTarget.objects.get(pk=obj.id)
        return ContentType.objects.get_for_model(base_target_instance)
    return ContentType.objects.get_for_model(obj)
```
But I don't know if that is a good idea. 
- in general, in our current setup regular users don't seem to be allowed to add targets to groups unless they have created the group and have the `change_target` permission for the target. We don't add that permission for groups in our ingest scripts, so I think no user has that permission for these targets, and this will always fail. If a target is created through the UI and permission is set to public, regular users seem to be able to add these targets to their own groups. But other users can only add these targets to their own groups if (I think) the user that created the target is in the same group as the one who wants to add it to their group.

Fingel:


jchate6 (changes requested):
This looks really good.
The issue I'm seeing is that it feels like this is still not quite the expected behavior.

I think this works really well when targetgroup creation is restricted. But if non-admin users are able to create groups, they should be able to add any target they have access to to that group. Basically, it seems like you need to have created a target to add it to a group you created. 
This seems like an unexpected behavior. If I've created the group, and there are public targets, I should be able to add them to my group without issue.

Fingel:
> This looks really good. The issue I'm seeing is that it feels like this is still not quite the expected behavior.
> 
> I think this works really well when targetgroup creation is restricted. But if non-admin users are able to create groups, they should be able to add any target they have access to to that group. Basically, it seems like you need to have created a target to add it to a group you created. This seems like an unexpected behavior. If I've created the group, and there are public targets, I should be able to add them to my group without issue.

This is what I tested:

I opened the TOM in 2 different windows, one logged in a user1 and another in incognito mode as user2.

I had user 2 create a target, and set it's permissions to public.

Back on the user 1 window, I reloaded the page, and saw the public target. I then created a group, and added the target to the group. I then confirmed that I could view that group and the only target it contained was the public  target created by user 2.

What am I missing here? This seems to match the expected behavior you describe.

image

jchate6:
Is user1 an admin?
The scenario I'm getting failure on is a non-admin user creating a target group, and trying to add a public target.

So if we have user1= admin, user2 = non-admin

- user1 makes public target
- user2 makes group, tries to add public target

Fingel:
> Is user1 an admin? The scenario I'm getting failure on is a non-admin user creating a target group, and trying to add a public target.
> 
> So if we have user1= admin, user2 = non-admin
> 
>     * user1 makes public target
> 
>     * user2 makes group, tries to add public target

OK I think I found the issue. The original code checked to make sure the user had permissions _on the target_ instead of permissions _on the group_. I changed the check to make sure the user has permissions for the group before adding any targets to it, which I think is more correct and less surprising. Also seems to fix this particular case as well.

tom_targets/groups.py
```diff
@@ -28,8 +28,8 @@ def add_all_to_grouping(filter_data, grouping_object, request):
         return
     for target_object in target_queryset:
         try:
-            if not request.user.has_perm('tom_targets.change_target', target_object):
```
Fingel, line 31: Here is the change from checking target permissions to checking targetlist permissions.

jchate6 (approved):
Excellent. Thank you!

PR #1555, opened 2026-05-13
Revealable password input partial

Fingel:


tom_common/templates/tom_common/partials/revealable_password_input.html
```diff
@@ -0,0 +1,11 @@
+<div style="cursor:pointer;"
+     title="Click to reveal"
+     onclick="var i=this.querySelector('input'); i.type = i.type === 'password' ? 'text':'password';">
+    <input
+        type="password"
+        readonly
+        value="{{ value }}"
+        class="form-control-plaintext p-0"
+        style="display:inline; width:auto; cursor:pointer;"
+    /> <i class="fa fa-eye"></i>
```
jchate6: Any easy way to force this to be in line?
Looks a bit awkward under the field. 

Image

Fingel: interesting, I thought it was. Maybe I need an "other stuff" field under to test it.

Fingel: Should be fixed in 2a86909c3e346a6066f79147a7dc7f75f72b99cd

jchate6 (approved):
beautiful

Issue #1461, opened 2026-03-19
UI sharing is broken


jchate6:
- [ ] #991 
- [ ] #1465

PR #1559, opened 2026-05-19
1485 remove broker code in v3

### LINKED ISSUE #1421: Management command needed to facilitate Dataservice queries.
see `tom_alerts.management.commands.runbrokerquery.py`

### LINKED ISSUE #1485: Remove Broker code in V3
- [x] remove tom_alerts as an installed app
- [x] Remove alerts from Nav Bar
- [ ] https://github.com/TOMToolkit/tom_base/issues/1421
- [ ] disconnect buttons on the Target List Page and Target Detail Page
  - [ ] move functionality to Dataservices
- [x] update settings.py template
- [x] Update docs

Bonus:
- [ ] add method to customize what dataservices show up in the nav bar dropdown
- [ ] update apps

jchate6:
This severs the broker code in tom_alerts from the rest of tom_base

24 of these files are docs changes. Mostly changing wording from alerts to dataservices

Currently we are not removing either `tom_catalogs` or `tom_alerts`. These apps will still be available if they are used by custom code. They should be considered deprecated though, and users should work on transferring any references to tom_dataservices. 

There is a little bit of complicated migration magic that needs to happen here. I think it should all be transparent and seamless as long as the `tom_alerts` app is removed from installed apps AFTER a migration is run for existing TOMs. The other situation is that someone has migrated tom_dataproducts migration 0011, but not 0013. This seems very unlikely though.

Instructions for testing this PR:

1. Fresh install: You're done. Migrate, then everything should work fine.
2. Install from TOM with an earlier version of V3 currently installed:
   - install this branch
   - add `tom_alerts` to list of installed apps
   - migrate
   - remove `tom_alerts` from list of installed apps
3. Install from some other, earlier version of the tomtoolkit:
   - install this branch
   - migrate
   - remove `tom_alerts` from list of installed apps


tom_alerts/tests/tests.py
jchate6, line 1: We are skipping all of the tom_alerts tests for now, so they don't pollute our test suite.

tom_dataproducts/migrations/0011_reduceddatum_message.py
jchate6, line 1: This change allows for this migration to be referenced by TOMS that have already run this migration.
New TOMs should use the squashed 11-15 migration below.

This migration should be able to be removed a reasonable time after release of V3 once TOMs have fully migrated.

tom_dataproducts/migrations/0011_reduceddatum_message_squashed_0015_remove_reduceddatum_message.py
jchate6, line 1: This squashed migration will replace migrations 11-15 for new TOMs and TOMs that have already completed all 5 migrations.

tom_dataproducts/migrations/0013_alter_reduceddatum_message.py
jchate6, line 1: same as 0011

tom_dataproducts/migrations/0015_remove_reduceddatum_message.py
jchate6, line 1: New migration to remove reference to tom_alerts.
Needs to be run before squashed migration applies if 0011-0014 has been run

tom_dataproducts/templates/tom_dataproducts/partials/update_broker_data_button.html
jchate6, line 1: Removed this functionality since it didn't work. Could be re-routed to use refactored updatereduceddata.py management command above.

tom_dataproducts/tests/test_sharing.py
jchate6, line 1: We should re-establish these tests when we fix sharing, but these only use Hermes at the moment.

jchate6: See #1461

tom_dataproducts/models.py
```diff
@@ -400,7 +400,6 @@ class ReducedDatum(models.Model):
     source_location = models.CharField(max_length=200, default='', blank=True)
     timestamp = models.DateTimeField(null=False, blank=False, default=timezone.now, db_index=True)
     value = models.JSONField(null=False, blank=False)
-    message = models.ManyToManyField(AlertStreamMessage, blank=True)
```
jchate6, line 403: This is the problem field that we removed.
This was used to track provenience of a datum that was retrieved via a hermes alert message.

tom_dataproducts/sharing.py
```diff
@@ -228,17 +140,11 @@ def get_destination_target(target, targets_url, headers, auth):
 def check_for_share_safe_datums(destination, reduced_datums, **kwargs):
```
Fingel: Is there a reason to keep this function around? Expand it in the future? Looks like it's a noop to me now.

tom_dataservices/dataservices.py
```diff
@@ -360,15 +361,16 @@ def to_target(self, target_result=None, **kwargs):
...
-                messages.warning(request,
-                                 mark_safe(
-                                    f"""The target,
-                                    <a href="{reverse('targets:detail', kwargs={'pk': target.id})}">
-                                    {target.name}</a> already exists, any new data has been ingested.
-                                    You can <a href="{reverse('targets:create') + '?' +
-                                                      urlencode(target.as_dict())}">create</a> a new target anyway.
-                                    """)
-                                 )
+                message = f"""The target,
```
Fingel: Minor nit: place message in the if block

tom_setup/templates/tom_setup/settings.tmpl
```diff
@@ -235,26 +235,6 @@ TOM_FACILITY_CLASSES = [
     'tom_observations.facilities.gemini.GEMFacility',
 ]
 
-TOM_ALERT_CLASSES = [
```
Fingel, line 238: 🥳

Fingel (approved):
Looks great overall! Just 2 really minor nits. Nice work!

docs/api/management_commands.rst
jchate6, line 1: Update this file

tom_dataproducts/sharing.py
jchate6: I believe the original intent here was to have this eventually become an integration point for users to have a method overriding this method that determined if some datum was OK to be shared.

The same is true for the `check_for_save_safe_datums` below.  I think the right move is probably to make an issue and remove this code.

docs/data_services/index.rst
```diff
@@ -0,0 +1,23 @@
...
+
+  create_dataservice
+  ../api/tom_dataservices/data_services
+  ../api/tom_dataservices/views
+
+
+What is a Data Service Module?
+-------------------------------
+
+A TOM Toolkit Data Service Module is an object which contains the logic for querying a remote broker or catalog service
```
phycodurus: How about: A TOM Toolkit Data Service Module ~is an object which~ contains the logic

docs/introduction/tomarchitecture.rst
```diff
@@ -184,19 +184,24 @@ or in the cloud) as well as displaying certain kinds of data. It also provides
...
 The
-`alerts.py <https://github.com/TOMToolkit/tom_base/blob/main/tom_alerts/alerts.py>`_
+`dataservices.py <https://github.com/TOMToolkit/tom_base/blob/main/tom_dataservices/dataservices.py>`_
 module provides a generic interface that other modules can implement, giving
-them the ability to integrate these brokers with the toolkit. Currently, there are
+them the ability to integrate new dataservices with the toolkit. Currently, there are
 modules available for `Lasair <https://lasair.roe.ac.uk>`_,
-`MARS <https://mars.lco.global>`_, `SCOUT <https://cneos.jpl.nasa.gov/scout/intro.html>`_, and others,
+`ANTARES <https://antares.noirlab.edu/loci>`_, `SCOUT <https://cneos.jpl.nasa.gov/scout/intro.html>`_, and others,
 with more planned for the future.
```
phycodurus: Fink?

phycodurus (approved):

PR #1561, opened 2026-05-21
Always show 10 hidden password characetrs on revelable password input

Fingel:


jchate6 (approved):
We hit our transient datetime issue, but I re-ran the tests and made an issue for it (#1562 ).
Looks good! thanks for this.

PR #1567, opened 2026-05-28
Add a default name to unnamed dataservice queries

### LINKED ISSUE #1564: Unnamed DataService query cumbersome
It's possible (for SIMBAD, at least) to save a query by checking the "Save query" box and not supplying a name. Then in the Saved Query ListView, the query shows up and is runnable, but in the Name column (where, normally a link to the query update form would appear (with the query name given), there is not text (or link). This makes it hard to edit the query.

We could make the QueryName field mandatory, default it to the target, or something else for a better UX.

Fingel:
Just adds [data service name] - [id of query] if a name isn't provided.


tom_dataservices/forms.py
jchate6, line 21: Any chance we can add some placeholder/help text describing this new behavior?

jchate6 (changes requested):
This looks great. One request.

jchate6 (approved):

PR #1571, opened 2026-06-04
1238 upgrade to bootstrap5

### LINKED ISSUE #1238: upgrade to bootstrap5
This will become an issue as more packages drop support for versions of Python that have reached end-of-life (see [python version status](https://devguide.python.org/versions/)).

Migration guides:

- [bootstrap4 to boostrap5](https://getbootstrap.com/docs/5.0/migration/)
- [django-bootstrap4 to django-bootstrap5](https://github.com/zostera/django-bootstrap5/blob/main/MIGRATE.md#:~:text=Migration%20Guide%20Below%20is%20a%20list%20of,Bootstrap%20docs%2C%20especially%20the%20Migration%20section%20_.)
- [crispy-bootstrap4 to crispy-bootstrap5](https://github.com/django-crispy-forms/crispy-bootstrap5)

This will require changes to several (many?) templates. If users have customized those templates, then this will complicate their upgrade paths.

jchate6:
Still very much a work in progress.

Still to be done:

- [x] bootstrap4 -> django_bootstrap5
- [x] replace `{% buttons %} ... {% endbuttons %}` tag with `[{% bootstrap_button %}](https://django-bootstrap5.readthedocs.io/en/latest/templatetags.html#bootstrap-button)`
- [x] update `CRISPY_TEMPLATE_PACK = 'bootstrap5'` and `CRISPY_ALLOWED_TEMPLATE_PACKS = "bootstrap5"` in settings.py
- [x] Fix `data-toggle` now `data-bs-toggle`
  - [x] data-dismiss
  - [x] data-target
- [x] Fix navbar text color
- [x] fix tab selection
- [x] fix new event calendar button
- [x] Remove `$` Jquery calls
- [x] target Select Button at bottom of page
- [x] `var(--pink);` --> `var(--bs-pink);`
- [x] Change `mr-` and `ml-` to `me-` and `ms-` respectively (for start/end instead of left/right)
- [x] `border-right/left` -> `border-end/start`
- [x] `font-weight` `font/style` -> `fw`, `fst`
- [x] Convert `form-row`, `form-group`, and `form-inline` to use grid/utilities
  - [x] `form-row` --> `row`
  - [x]  `form-group` --> `mb-3`
  - [x]  `form-inline` --> `d-flex flex-row align-items-center flex-wrap`

Note: link buttons:>
```
{% url 'tom_dataproducts:list' as reset_link %}
{% bootstrap_button "Reset" button_type="link" button_class="btn btn-secondary" href=reset_link title="Reset" %}

```


docs/customization/testing_toms.rst
```diff
@@ -139,8 +139,8 @@ option requires two short files in addition:
         'guardian',
         'tom_common',
         'django_comments',
-        'bootstrap4',
-        'crispy_bootstrap4',
+        'bootstrap5',
```
Fingel: My TOM won't load unless this is `django_bootstrap5`

tom_base/settings.py
```diff
@@ -68,7 +68,7 @@
     },
 ]
 
-CRISPY_TEMPLATE_PACK = 'bootstrap4'
+CRISPY_TEMPLATE_PACK = 'bootstrap5'
```
Fingel, line 71: I also need this:

`CRISPY_ALLOWED_TEMPLATE_PACKS = "bootstrap5"`

Otherwise I get this one some pages:

TemplateSyntaxError at /targets/

crispy tag's template_pack argument should be in ('uni_form', 'bootstrap3', 'bootstrap4')

for example, /targets

tom_common/default_settings.py
Fingel, line 1: Ah, these are the correct settings values I commented on earlier. I guess they just need to be made consistent.

Fingel (approved):
Looks good, much cleaner visually now.

docs/customization/testing_toms.rst
jchate6: good catch.

tom_base/settings.py
jchate6: I did include that in `tom_common/default_settings`
which `tom_base.settings` imports. 
Hopefully you don't have those errors if you import the default settings or use the `settings.tmpl` when setting up your local TOM?

jchate6: I'm also open to the idea that default settings is not the place for this, but it seemed like the easiest way for users to get the change without having to actually do anything.
I could also see moving this line `CRISPY_TEMPLATE_PACK = 'bootstrap5'` into default settings as well, or pulling the CRISPY_ALLOWED_TEMPLATE_PACKS out and just adding it to the settings update list required for V3. Thoughts?

tom_common/default_settings.py
jchate6: lol

PR #1581, opened 2026-06-17
Scope reveal script to closest element, fixes #1578

### LINKED ISSUE #1578: New encryption functionality is broken
Setting and clearing encrypted fields in the update profile form of a profile appears to clear or sets ALL above encrypted fields as well. 

Basically, if you install 2 apps with encrypted fields, the first field will always display the same value as the lower field, regardless of what is actually stored there. So revealing the password on the profile will always show the value stored in the 2nd box.

Fingel:


jchate6 (approved):
Excellent! thank you!

PR #1588, opened 2026-06-24
1586 custom target sharing

### LINKED ISSUE #1586: How does TOM-TOM sharing behave with custom Target models?
Two TOM teams would like to share Targets and data between their TOM-Toolkit TOM systems.  At least one of the teams has customized their Target model with science-specific attributes. 

- [x] What happens to the custom attributes when Targets are shared from one TOM to another?
- [x] Do they need to add a custom sharing function to include the specific attributes?
- [x] Please extend the [TOM-TOM sharing documentation](https://tom-toolkit.readthedocs.io/en/3.0.0-alpha11/managing_data/tom_direct_sharing.html) to address this question.

Fingel:
When sharing, converts custom target model fields to target extras for the destination tom by comparing the custom model ._meta.fields() with the BaseTarget._meta.fields()

Two other fixes:

2279497214b2e303d5b95f43eb6fbcb4a908e24b Sharing was broken for custom targets to begin with due to mismatched app labels

de441311fc4aa20b40e100204a53ecfffc74f9df fix submit button being placed at the bottom of the screen far away from everything

Fixes #1586


jchate6 (changes requested):
testing this, I'm getting a `"POST /api/targets/ HTTP/1.1" 400 64` in the receiving TOM, and `ERROR: Returned Response code 400 with content: b'{"targetextra_set":[{"value":["Not a valid string."]},{},{},{}]}'` as an error message in my sending TOM that has a custom Target. 
I have several additional fields in my extended target model.
Thoughts?

Fingel:
I couldn't reproduce this with additional custom fields. Could this be an authentication issue masking the actual sharing? Do you have permissions with the auth key you are using for the user on Tom B?

In any case, I added a string cast to the target extra value, though it should work with values of other types.

jchate6:
ok, I'll try to figure out what' going on.

tom_targets/sharing.py
```diff
@@ -37,6 +38,20 @@ def continuous_share_data(target, reduced_datums):
         share_data_with_tom(share_destination, None, None, None, selected_data=reduced_datum_pks)
 
 
+def custom_target_to_extras(target_id) -> list[dict]:
+    target_app_label = get_target_model_app_label()
+    extra_fields = []
+    if target_app_label != 'tom_targets':
+        target = get_target_model_class().objects.get(pk=target_id)
+        for field in target._meta.get_fields():
+            if field not in BaseTarget._meta.get_fields() and field.name not in ['id', 'basetarget_ptr']:
```
jchate6, line 47: I believe this assumes that the receiving TOM doesn't have a custom Target Model.
If the receiving TOM also has an updated model, and that model includes some of the same fields as the sending TOM, then this will create both a proper model entry AND an extra field:

Image

jchate6 (changes requested):
Starting over, this time I didn't see the error from before, but did find a small bug.

PR #1589, opened 2026-06-24
1587 fix issue with simple fields on update form

### LINKED ISSUE #1587: Fix issue with simple fields on update form

jchate6:
There was some missing context from the update form for a dataservice query

This has escalated to be a more important update necessary for tom_hermes


tom_dataservices/views.py
```diff
@@ -321,12 +321,24 @@ def get_context_data(self, *args, **kwargs):
...
+        data_service_name = self.object.data_service
 
-        simple_form = context['form'].get_simple_form_partial()
-        advanced_form = context['form'].get_advanced_form_partial()
+        form = context['form']
+        simple_form = form.get_simple_form_partial()
+        advanced_form = form.get_advanced_form_partial()
+
+        context['simple_fields'] = []
+        if not simple_form and form.simple_fields():
```
Fingel, line 336: Why would the form not have as simple partial, but also have simple fields?

I trust that this works, but it just reads a little strange.

Fingel (approved):

tom_dataservices/views.py
jchate6: I describe this in the [docs](https://tom-toolkit.readthedocs.io/en/latest/data_services/create_dataservice.html#simple-forms) but I should definitely add a comment here.
Basically, there is a basic partial that takes the fields in `simple_fields` and makes a super basic form. OR you can can make a simple partial that does whatever you want. So if no partial is given, but there are simple fields we will just use those in the built in default partial.

jchate6: One would not necessarily expect there to be BOTH a `simple_fields` and a user made `simple_partial`, since the user made simple partial could be much more specific to the dataservice's form fields.

docs/data_services/create_dataservice.rst
```diff
@@ -357,16 +390,16 @@ exists, in several places where we want to update an existing target based on da
...
-                form_fields = {'first_field': target.name}
-                query_parameters = self.build_query_parameters(form_fields)
-            else:
-                query_parameters= {
-                    'ra_field': target.ra,
-                    'dec_field': target.dec,
-                    'radius': 0.5
-                    }
-            return query_parameters
+        if 'first' in target.name:
```
Fingel, line 394: What is this special case when a target name contains 'first'? Seems worthy of a comment.

Fingel (approved):
Looks good, just one comment on the 'first' thing.

tom_dataservices/views.py
```diff
@@ -175,19 +177,20 @@ def get_context_data(self, *args, **kwargs):
...
                 query = get_object_or_404(DataServiceQuery, pk=self.kwargs['pk'])
-                data_service_class = get_data_service_class(query.data_service)()
+                data_service_class = get_data_service_class(query.data_service)(user=user)
                 query_parameters = data_service_class.build_query_parameters(query.parameters)
                 query.last_run = timezone.now()
                 query.save()
             else:
                 input_parameters = self.request.session.get('query_parameters', {})
-                data_service_class = get_data_service_class(input_parameters['data_service'])()
+                data_service_class = get_data_service_class(input_parameters['data_service'])(user=user)
```
phycodurus: (Also on line 187): Aren't you instantiating the data service class here? So, the variable is named `data_service_class` (inplying that it's value is the _class_) but it's value is a _data service_. 

TL;DR: I think this variable should be `data_service` not `data_service_class`.

phycodurus (approved):

Issue #1597, opened 2026-07-06
Target sharing questions


rachel3834:
May want to limit user’s ability to share targets.   Record who shared the target.  What happens with two-way sharing?

**Is your feature request related to a problem? Please describe.**
Two microlensing groups with TOM-Toolkit-TOMs are experimenting with sharing individual targets.  The option  to persistently share target data is appreciated, but applying this in practise has led to the following questions:

- [ ] #1604
- [x] #1605
- [x] #1606

**Describe the solution you'd like**
Items 1 and 2 would be highly beneficial and are desired for LSST follow-up, i.e. as soon as possible. 
Item 3 is understood to be more complicated and custom APIs could be written if needed.

PR #1603, opened 2026-07-08
Track user as target extra when sharing

### LINKED ISSUE #1605: Track Target Provenance
In the interests of tracebility, it is desirable to record the user ID and origin-TOM who shared the target.

Fingel:
adds a "shared_by" target extra when sharing a target. This is part of item 2 in #1597


Fingel:
Change this to use model fields directly instead of target_extras

Fingel:
shared_by and shared_from are now added directly to the base target model

jchate6 (approved):
This looks good, but would benefit from a test or 2. 

Also, I ran into an issue with a 
`sqlite3.IntegrityError: UNIQUE constraint failed: tom_targets_targetextra.id`
When trying to share from TOM1 to TOM2 with a target that had been deleted in TOM2.

Hilariously, the entire stack trace was piped into the error message in TOM1

Image

I don't think this is related to this code, but was hoping you could take a look just in case and either apply a quick fix, or make an issue if you can reproduce it?

jchate6 (approved):

Fingel:
I can't reproduce this issue.
I have 2 TOMs:

From TOM A, I share target a1.
From TOM B, I reload and see target a1.
From TOM B, I delete target a1.
From TOM A I share target a1 again.
From TOM B, target a1 is available again, but with a new ID.

This is what is expected. Are there concrete steps I can take to cause this exception to occur? The error itself is strange because IDs should be assigned automatically, this is some query specifying an ID to save a TargetExtra as, but I can't find any code that would do that.

jchate6:
Those are the steps I used to get that error. It's possible something else is going on? Are you using a postgres or sqlite DB?

I'm leaning towards just merging this. If it comes up again we can make an issue.

Fingel:
> Those are the steps I used to get that error. It's possible something else is going on? Are you using a postgres or sqlite DB?
> 

Using sqlite for both TOMs

> I'm leaning towards just merging this. If it comes up again we can make an issue.

It's pretty hard to see how this PR specifically would cause the issue. Not saying there wasn't something introduced in the larger sharing work that has been done, but this is really piping 2 values over.

PR #1607, opened 2026-07-09
Makes sure sharing of reduceddatums/dataproducts doesn't fail if reci…

### LINKED ISSUE #1606: Add option to retrieve new data from TOM when sharing Target with TOM
Is it possible to have two-way sharing? E.g. one TOM shares the Target details and some data products with a second TOM, and the second TOM shares additional data products back?

Fingel:
…eving side already has them


jchate6 (approved):
This seems fine, but doesn't keep you from uploading the same DataProduct multiple times. 

At least locally, the file system just modifies the string so it has a different name.

Image

I'm happy to leave this to the User to sort if we want, but I thought it should be pointed out.

Fingel:
I kinda like that rename behavior.

PR #1610, opened 2026-07-10
1609 fix button group behavior

### LINKED ISSUE #1609: Fix Button Group behavior
The Hermes Target Detail button is inside a form object, and therefore not recognized by the bootstrap [button group](https://getbootstrap.com/docs/5.0/components/button-group/) object.

This is done to allow for a post to the view, but this means that the corners of the button are not correct:

Image
vs.
Image

jchate6:
image

Feels not perfect, but better....

### DIFF
--- a/tom_targets/templates/tom_targets/partials/target_buttons.html
+++ b/tom_targets/templates/tom_targets/partials/target_buttons.html
@@ -1,5 +1,5 @@
 {% load targets_extras %}
-<div class="btn-group" role="group" aria-label="Target detail buttons">
+<div role="group" aria-label="Target detail buttons">
     <a href="{% url 'tom_targets:update' pk=target.id %}" title="Update target" class="btn  btn-primary">Update</a>
     {% if sharing %}
     <a href="{% url 'tom_targets:share' pk=target.id %}" title="Share target" class="btn  btn-info">Share</a>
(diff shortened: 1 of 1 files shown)


phycodurus (changes requested):
How about instead of this, in `tom_hermes/templates/tom_hermes/partials/hermes_share_button.html`, we separate the form from the button, then let bootstrap5 CSS just treat the button like normal?

This is what `tom_hermes/templates/tom_hermes/partials/hermes_share_button.html` would become:
```html
{# Separate the form from the button for CSS styling compatibility. #}
{# Use form attr in button below #}
<form id="hermes-share-form" method="post"
      action="{% url 'tom_hermes:target-preload' pk=target.id %}" class="d-none">
  {% csrf_token %}
  <input type="hidden" name="share_destination" value="{{ hermes_topic }}">
  <input type="hidden" name="submitter" value="{{ request.user }}">
</form>

{# Now the button can be a part of the button group and get the standard styling #}
<button type="submit" form="hermes-share-form" title="{{ button_text }}"
        class="btn btn-info">{{ button_text }}</button>
```

jchate6:
> How about instead of this, in `tom_hermes/templates/tom_hermes/partials/hermes_share_button.html`, we separate the form from the button, then let bootstrap5 CSS just treat the button like normal?

That was my first thought, but there are other issues with the button group. 
First, if the button text is several words, it just makes the whole group really fat in a less than appealing way.  
2nd if you have too many buttons they overflow:
image

This solution maximizes utility and flexibility. I would like to add some padding between rows, so feel free to spend some time fixing that if you like, but I think it's a good thing to not restrict the button format from being inside a form, and have reasonable behavior for long, or unusual button text.

image

jchate6:
image

Updated margins for child objects

phycodurus (approved):

PR #1615, opened 2026-07-13
Encryption documentation updates

phycodurus:


jchate6 (changes requested):
These are fine, but the issue is a separate page describing GENERAL security and encryption when using a TOM Toolkit. 

This is just additional (probably unnecessary) information about the backend encryption for this one specific type of field. 

I think most of this clutters the existing docs which are there to describe HOW to implement this feature, rather than the specifics of it inner workings. I think it's still OK and readable, but not what most users would be looking for here.

AGAIN, the big problem is that the issue represents a request for a new section in the doc describing TOMToolkit Security GENERALLY. This would be things like information on DB security, Django password management, Links to our encrypted field documentation, and TOM_registration, etc.

phycodurus:
@jchate6 I suggest we approve and merge this PR and then re-open the original issue, adding a description for what you want. How does that sound?

jchate6 (approved):

PR #1618, opened 2026-07-16
Prevent Aladin chart from blocking page load

### LINKED ISSUE #1617: Optimize Aladin loading
Loading Aladin from Strasbourg slows down page loads. Remedies include:
- [ ] cache aladin
- [ ] package aladin with TOMToolkit static files
- [ ] Load aladin asynchronously

Fingel:
Moves the javascript into a self contained file, so that it can be loaded in the {% extra_javascaript %} block which is outside of \<body\>, thus preventing it from blocking the loading of the rest of the page.

Required some refactoring as the original JS used template variables, which needed to be converted to parameters for the standalone function.

### DIFF
--- a/tom_targets/static/tom_targets/js/aladin_finderchart.js
+++ b/tom_targets/static/tom_targets/js/aladin_finderchart.js
@@ -1,0 +1,173 @@
+(function () {
+  'use strict';
+
+  function initAladin(targetRa, targetDec) {
+    let aladin;
+    A.init.then(() => {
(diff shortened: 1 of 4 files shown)


jchate6 (approved):
This is great. A few questions:

- [ ] Can we make it so there is a spinning wheel or equivalent while loading the finder?
- [ ] The finder still feels very slow. the main aladin star chart on the Target Details page is much faster. Is there anything we can do to speed up the individual target finder chart?

PR #1624, opened 2026-07-17
Remove special unique constraints ReducedDatums

### LINKED ISSUE #1622: Dataservice target creation issue
- [ ] Target creation page from Alerce -- ZTF21abrotmc
  -  IntegrityError at /tom_dataservices/query/create_targets/  -- ```duplicate key value violates unique constraint "unique_photometry"
DETAIL:  Key (target_id, bandpass, "timestamp")=(8, g, 2021-08-14 06:33:19.002233+00) already exists.```

jchate6:
The unique constraints we used to use were necessary because validating on the "value" field wasn't possible. This is no longer the case.


jchate6:
@Fingel Without these constraints, I think we need to add some check in the API loop to make sure we aren't ingesting identical data.

I'm not sure where that should be... `serializers.ReducedDatumSerializer.create`? `models.try_parse_reduced_datum`?

Maybe we just need to include more fields in the constraints, rather than remove them, but this feels like it can be solved by a `get_or_create` somewhere rather than a constraint, since we have actual known fields we can test against now.

Suggestions?

Fingel:
I'm always going to prefer database constraints to (potentially) faulty business logic. A constraint doesn't prevent a get_or_create() from working either, they aren't mutually exclusive.

I think what actually happened here is the use case of duplicate reductions wasn't taken into account. I think Rachel was correct when she said that a version number should be tracked and added to the constraint, or alternatively one could use an upsert pattern to either update the incoming datum, or create a new one if it doesn't exist.

I don't think removing DB constraints entirely is a good idea.

jchate6:
@Fingel Ready for another review.
Added error handling to catch issue, and added values to uniqueness constraint.

Can you help me understand what's going on with the API?
It seems to behave the same as before with the value field JSON

image

kwynn03:
I've added a field for 'reduction version' in the value JSON (containing the md5 hash) in the app I'm working on, but perhaps having it as its own field in the model would help to stop duplicate entries. I also have a field naming the 'reducer' for each ReducedDatum, but that wouldn't help with the uniqueness issue.

tom_dataproducts/migrations/0018_remove_astrometryreduceddatum_unique_astrometry_and_more.py
```diff
@@ -0,0 +1,58 @@
...
+
+    dependencies = [
+        ('tom_dataproducts', '0017_spectroscopyreduceddatum_wavelength_unit'),
+        ('tom_targets', '0030_alter_basetarget_slope'),
+    ]
+
+    operations = [
+        migrations.RemoveConstraint(
+            model_name='astrometryreduceddatum',
+            name='unique_astrometry',
```
Fingel, line 16: This appears to remove the constraint.

tom_dataproducts/models.py
```diff
@@ -453,10 +458,16 @@ class PhotometryReducedDatum(ReducedDatumCommon):
     class Meta:
         constraints = [
             models.UniqueConstraint(
-                fields=["target", "bandpass", "timestamp"], name="unique_photometry"
+                fields=["target", "bandpass", "timestamp", "limit", "brightness", "instrument", "reduction_version"],
+                name="unique_photometry",
```
Fingel, line 462: But the constraint is still present here?

Fingel (changes requested):
Little confused about what this PR does. It looks like it removes the constraints from the database, but they haven't been removed from the models. In fact it looks like some of the model constraints have had fields added.

tom_dataproducts/migrations/0018_remove_astrometryreduceddatum_unique_astrometry_and_more.py
jchate6: So, I'm not sure why CHANGING the constraint, removes the constraint and then re-applies it in the migration (see line 46). But that seems to be how it works for constraints in django. 

Also adding to confusion is that removing the constraint was the original goal of this PR, but after our discussion I changed tack.

PR #1639, opened 2026-08-05
Added tom_across to the plugins page

moira-andrews:
I added a description of the TOM across app to the `plugins.rst` page. Feel free to request any rephrasing or move it from the top of the page. I wasn't sure the ordering of the plugin list.


jchate6 (approved):
This is fine, but I'd like non-sidereal targets fixed before we advertise tom_across in the docs.
See:
-  https://github.com/TOMToolkit/tom_across/issues/33

phycodurus (approved):

PR #1648, opened 2026-08-18
fix datetime format for Django 5

griffin-h:
Django 5 dropped support for specifying date formats in the settings file. This is the recommended workaround, e.g., https://www.loopwerk.io/articles/2025/django-i18n-date-formats/.


phycodurus (approved):

Fingel (commented):
This will only be applied to the en locale. Which might be fine. I'm curious, is there a way to override all locale's formats to the one we were using? I think that would be consistent with the desired behavior from before.

griffin-h:
> This will only be applied to the en locale. Which might be fine. I'm curious, is there a way to override all locale's formats to the one we were using? I think that would be consistent with the desired behavior from before.

This is a good point, but I have no idea how to do it!

Fingel:
> This is a good point, but I have no idea how to do it!

I did some digging myself and I don't think there is a way. I think we just accept that if the TOM developer changes the locale to something other than en-us then they get whatever datetime format is set for their locale. Which might be the desired outcome, anyway. Though I assume the defaults for other languages are just as "unscientific" as the one we are overriding here.

The real fix here is not to place bare {{ times }} in templates and explicitly format them with a constant, defined in settings. So like DATETIME_FORMAT but called something else so we aren't shadowing a deprecated setting. That's a whole other issue and chunk of work though. I think this is fine.

Fingel (approved):

PR #1651, opened 2026-08-26
enable get_latest_by on *ReducedDatum

griffin-h:
Previously, the `ReducedDatum` model had a `Meta.get_latest_by` attribute, enabling queries like `ReducedDatum.objects.latest()`. This has been lost in the specific `*ReducedDatum` models. This model moves the specification to the `ReducedDatumCommon` class and inherits it in the child models.


Fingel (approved):
Thank you!

Issue #1405, opened 2026-02-10
Babamul alert broker data service


rachel3834:
**Is your feature request related to a problem? Please describe.**
Babamul is one of the LSST and ZTF alert brokers, which has recently announced a new public service, which offers an API:
https://babamul.caltech.edu/

**Describe the solution you'd like**
The toolkit should offer a dataservice for users to query this service.

DKilkenny:
Hello, 

I saw this was still open and thought it would be a good initial issue to help contribute to. Babamul has self-serve sign-up now, so I believe this is unblocked. You can sign up and create an API key to access it. Happy to answer any questions

PR #1666, opened 2026-09-10
Add Babamul alert broker DataService

DKilkenny:
Adds a DataService for Babamul, the public service of the BOOM alert broker for ZTF and LSST alerts.

Supports cone search and lookup by object id. Photometry is read from the object endpoint, which carries the full light curve: detections are stored as `brightness` and `prv_nondetections` as limits.

The API key is read from DATA_SERVICES, following the TNS pattern.

Refs issue #1405


DKilkenny:
I have two questions:

1. Every detection I'm seeing carries `isdiffpos: false` and I'm currently ingesting all of them. The way I understand it is that a negative subtraction usually gets excluded from a light curve, so if that's the convention, I should be filtering them out.
2. How much enrichment do you want surfaced? Right now the only extra I save is `survey` since it decides which endpoint a later photometry update queries. I noticed the services here have two approaches. 
- ALeRCE stores every field that isn't on the Target model
- MPC, NED, Simbad and TNS store nothing at all
The object record has a lot available if you want it including ACAI and BTSbot classification scores and boolean flags. I chose the ZTF option for now, but I'm happy to implement the ALeRCE pattern instead.

Fingel (approved):
I tested this locally. Searching by name works, ra/dec works, saving the query works and creating targets. Looks good to me!

PR #1671, opened 2026-09-11
fix home URL

griffin-h:
This fixes the URL of the home button in the navbar for sites (like ours) that are not hosted at the root of their domain.


jchate6 (approved):
I've added a base.html change to do this as well.

PR #1675, opened 2026-09-16
Silence warning about unique constraints and null

### LINKED ISSUE #1668: New warnings about ReducedDatum uniqueness constraints
With a PostgreSQL database, I'm getting these warnings about the new *ReducedDatum models after upgrading to v3:
```
tom_dataproducts.AstrometryReducedDatum: (models.W047) PostgreSQL does not support unique constraints with nulls distinct.
	HINT: A constraint won't be created. Silence this warning if you don't care about it.
tom_dataproducts.PhotometryReducedDatum: (models.W047) PostgreSQL does not support unique constraints with nulls distinct.
	HINT: A constraint won't be created. Silence this warning if you don't care about it.
tom_dataproducts.SpectroscopyReducedDatum: (models.W047) PostgreSQL does not support unique constraints with nulls distinct.
	HINT: A constraint won't be created. Silence this warning if you don't care about it.
```
I don't know if this will be a problem, but I thought I would flag it here in case this causes unexpected behavior. Feel free to close if you investigate and find it's not a problem.

Fingel:
The reasoning behind using nulls_distinct=False is sound for reducceddatums, and the behavior is consistent between databases so this is OK.

Fixes #1668


jchate6 (approved):
nice

griffin-h:
I'm not sure this is a good idea, because it was these warnings that pointed out to us that we needed to upgrade to PostgreSQL>=15. If a user is still on PostgreSQL 14 or lower, the distinct constraint will now silently not be created.

jchate6:
@griffin-h 
What Tomtoolkit version are you using?
Also, can you explain what you meant by the uniqueness constraint not being included?

The behavior we were seeing was that, previously, 2 identical PhotometryReducedDatums could co-exist if the `limit` was `None`, since by default (for most DBs) `None`s are treated as distinct. This change SHOULD tell relevant DBs to treat `None`s as the same, and fail a uniqueness check if everything else is the same too. Are you saying that with PostgreSQL 14, with the changes after tomtoolkit 3.0.1, 2 identical datums could be added to the DB if `limit=None`?

Fingel:
I should have done this in the pull request itself, but I added this test:
https://github.com/TOMToolkit/tom_base/pull/1676/changes

It passes on both sqlite and postgresql. I guess the question is would it pass on Postgresql 14?

griffin-h:
> The reasoning behind using nulls_distinct=False is sound for reducceddatums, and the behavior is consistent between databases so this is OK.
> 
> Fixes #1668

I totally agree with your logic; I think that `nulls_distinct=False` should be in the code. It just isn't supported on our 5-year-old SQL server.

> @griffin-h What Tomtoolkit version are you using? Also, can you explain what you meant by the uniqueness constraint not being included?
> 
> The behavior we were seeing was that, previously, 2 identical PhotometryReducedDatums could co-exist if the `limit` was `None`, since by default (for most DBs) `None`s are treated as distinct. This change SHOULD tell relevant DBs to treat `None`s as the same, and fail a uniqueness check if everything else is the same too. Are you saying that with PostgreSQL 14, with the changes after tomtoolkit 3.0.1, 2 identical datums could be added to the DB if `limit=None`?

We're on 3.0.1, and we currently have a lot of duplicate photometry in our database because we have reduction_version=NULL for everything.

> I should have done this in the pull request itself, but I added this test: https://github.com/TOMToolkit/tom_base/pull/1676/changes
> 
> It passes on both sqlite and postgresql. I guess the question is would it pass on Postgresql 14?

I think the answer is it would not pass on PostgreSQL 14. It's not just a useless warning, it's actually telling us that the constraint is not created on our database.

Fingel:
The strange thing is that we get the same exact warning on sqlite currently (not an old version) but the test still passes. Here's the output of running the new test I added, with sqlite, with the suppression turned off:

```
❯ ./manage.py test --exclude-tag=canary tom_dataproducts.tests.tests.TestReducedDatumModel.test_create_reduced_datum_duplicate_none
Found 1 test(s).
Creating test database for alias 'default'...
System check identified some issues:

WARNINGS:
tom_dataproducts.AstrometryReducedDatum: (models.W047) SQLite does not support unique constraints with nulls distinct.
        HINT: A constraint won't be created. Silence this warning if you don't care about it.
tom_dataproducts.PhotometryReducedDatum: (models.W047) SQLite does not support unique constraints with nulls distinct.
        HINT: A constraint won't be created. Silence this warning if you don't care about it.
tom_dataproducts.SpectroscopyReducedDatum: (models.W047) SQLite does not support unique constraints with nulls distinct.
        HINT: A constraint won't be created. Silence this warning if you don't care about it.

System check identified 3 issues (0 silenced).
.
----------------------------------------------------------------------
Ran 1 test in 0.009s

OK
Destroying test database for alias 'default'...
```
I would expect this test to fail if the uniqueness check isn't happening, which is why I'm confused.

griffin-h:
I agree that is confusing. For the record, I tried to run the test on my own database, but I got an unrelated error about a custom app table not existing:
```
(tt3) griffin@ghosseinzadeh-laptop saguaro_tom % ./manage.py test --exclude-tag=canary tom_dataproducts.tests.tests.TestReducedDatumModel.test_create_reduced_datum_duplicate_none
Found 1 test(s).
Creating test database for alias 'default'...
Traceback (most recent call last):
  File "/Users/griffin/PycharmProjects/saguaro_tom/tt3/lib/python3.10/site-packages/django/db/backends/utils.py", line 103, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.UndefinedTable: relation "tom_surveys_surveyfield_adjacent" does not exist
```

jchate6:
> we currently have a lot of duplicate photometry in our database because we have reduction_version=NULL for everything.

I don't understand this. you have the same target/brightness/timestamp and `reduction_version=NULL`? As in you haven't pulled reduction_version out of the JSON blob yet? The uniqueness constraints would not apply to existing data already in the DB. It would just prevent new data from being added... with the PostgreSQL 14, were you able to add a new datum identical to the old ones? 

We can also chat about this tomorrow, if you want.

jchate6:
Results of discussion in 
- #1678

PR #1676, opened 2026-09-16
Modify ReducedDatum constraints to no longer require nulls_distinct=False

Fingel:
Reference test for issue #1668


jchate6 (approved):
I didn't think about this before, but we should consider making the settings change in default_settings rather than in settings.py and the template.

jchate6 (approved):
Much better

jchate6:
@griffin-h may want to have another look.

tom_dataproducts/models.py
```diff
@@ -456,9 +456,12 @@ class PhotometryReducedDatum(ReducedDatumCommon):
...
             models.UniqueConstraint(
-                fields=["target", "bandpass", "timestamp", "limit", "brightness", "instrument", "reduction_version"],
-                name="unique_photometry",
-                nulls_distinct=False
+                fields=["target", "bandpass", "timestamp", "brightness", "instrument", "reduction_version"],
+                name="unique_photometry_brightness",
+            ),
+            models.UniqueConstraint(
+                fields=["target", "bandpass", "timestamp", "limit", "instrument", "reduction_version"],
+                name="unique_photometry_limit",
```
griffin-h: Would this handle the situation where we have two photometry points with different brightnesses but the same limit? Whenever a limiting magnitude is available, we have been storing that in our database, even if the photometry point is a detection (i.e., it has a brightness). I think we would want two PhotometryReducedDatums to be allowed as long as the brightness OR the limit (or both) were different.

griffin-h (commented):
Just one small question above. And to make sure I understand, the idea is to merge this *instead* of #1668?

jchate6:
> Just one small question above. And to make sure I understand, the idea is to merge this _instead_ of #1668?

Technically #1675 has already been merged, but not released. This would undo those changes.

tom_dataproducts/models.py
jchate6: Speaking to @griffin-h I think we determined that this would not cause problems in realistic scenarios.

PR #1682, opened 2026-09-21
add email configuration docs

### LINKED ISSUE #1681: document-email-configuration
Email is a Django feature and there are no TOM Toolkit email configuration settings. However, TOM Toolkit documentation should have an `Email configuration` section that works together with Django's documentation.

phycodurus:
Adds email configuration docs. This was factored out of the MFA PR and ideally will merge into dev first.

closes #1681


docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
+Email Configuration
+===================
+
+When configured, TOM Toolkit sends emails for registration requests,
+registration approval notices, and password resets.
+
+Email delivery is a Django feature. There are no TOM Toolkit-specific email settings.
+See Django's `Sending email <https://docs.djangoproject.com/en/5.2/topics/email/>`_
```
jchate6: Change this link to "stable" so it doesn't fall out of date.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+Unless you've already done some
+configuration, that didn't work. Try this::
+
+   ./manage.py sendtestemail --help
+
+Notice the ``--managers`` and ``--admins`` options and their references to ``settings.MANAGERS`` and
+``settings.ADMINS``, respectively.
+
+Let's configure those in your ``settings.py``. We'll configure an ``EMAIL_BACKEND`` while
+we're at it)::
```
jchate6: you don't open the parenthetical

jchate6: Up to you if you want to include the parenthetical at all. I think a period would be just fine here.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+  Subject: [Django] Test email from tomtoolkit-host on 2026-09-19
+   00:32:32.742613+00:00
+  From: root@localhost
+  To: mary@example.com
+  Date: Sat, 19 Sep 2026 00:32:32 -0000
+  Message-ID: <178977795274.2655038.15656462345865752721@tomtoolkit-host>
+  
+  This email was sent to the site managers.
+  -------------------------------------------------------------------------------
+
```
jchate6: A conclusion to this section would be nice.
Recap what we've learned, re-state important points and say what we will be doing next.
Like a few sentences at most.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+(See :doc:`Accounts and Authentication <authentication>` for the features themselves.)
+
+When email is not configured, or a send fails, these features degrade with guidance in the UI
+rather than breaking (see `What happens when email is not configured or  when sends fail`_ below).
+
+Configuring the backend
+-------------------------
+
+Email delivery is standard Django — TOM Toolkit adds no email settings of its own. The complete
```
jchate6: These two sentences are poorly written.
```
Here we discuss how to configure email delivery within your TOM. It should be noted that all of this functionality is established by Django, and none of these settings are unique to the TOMToolkit. Any details that are not covered here can be found in the `Django Docs<https://docs.djangoproject.com/en/5.2/topics/email/>`_ .
```

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    EMAIL_HOST = 'smtp.example.org'         # your institution's or provider's SMTP relay
+    EMAIL_PORT = 587
+    EMAIL_HOST_USER = 'tom@example.org'
+    EMAIL_HOST_PASSWORD = os.getenv('EMAIL_HOST_PASSWORD', '')  # keep secrets out of settings.py
+    EMAIL_USE_TLS = True
+    DEFAULT_FROM_EMAIL = 'tom@example.org'  # the From: address on everything the TOM sends
+    MANAGERS = [('TOM admins', 'admins@example.org')]  # registration requests go here
+
+The host, port, credentials and TLS mode come from your email provider; what each setting means is
```
jchate6: "...from your email provider. See Django's `email settings reference <https://docs.djangoproject.com/en/5.2/ref/settings/#email-backend>`_ for more information about individual settings."

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+Configuring the backend
+-------------------------
+
+Email delivery is standard Django — TOM Toolkit adds no email settings of its own. The complete
+reference is Django's `Sending email <https://docs.djangoproject.com/en/5.2/topics/email/>`_
+topic guide; this section only puts the settings in TOM context.
+
+In the tutorial section above, we configured an email backend that prints to stdout, which is useful
+for development. For production, point Django's SMTP backend (the default) at your mail relay, and
+say who your TOM's mail comes from and who its administrators are::
```
jchate6: You mention admins, but only include managers in these settings. 
Above you include both.
I think this is a potentially confusing point, and we need to be careful about what we mean with each.
For Django, managers and Admin are treated differently.
A Note explaining the difference might be a good idea.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+
+    EMAIL_HOST = 'smtp.example.org'         # your institution's or provider's SMTP relay
+    EMAIL_PORT = 587
+    EMAIL_HOST_USER = 'tom@example.org'
+    EMAIL_HOST_PASSWORD = os.getenv('EMAIL_HOST_PASSWORD', '')  # keep secrets out of settings.py
+    EMAIL_USE_TLS = True
+    DEFAULT_FROM_EMAIL = 'tom@example.org'  # the From: address on everything the TOM sends
+    MANAGERS = [('TOM admins', 'admins@example.org')]  # registration requests go here
```
jchate6: I think including `TOM admins` here in the managers list is confusing when there is an `ADMINS` setting in django.
I'd rename this to TOM Managers or something to represent that these people approve registration, not necessarily handle code for the TOM.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Verifying your configuration
+----------------------------
+
+As seen in the tutorial, Django includes a management command that sends a test message
+through whatever backend you configured::
+
+    ./manage.py sendtestemail you@example.org
+
+To try the SMTP configuration without involving a real relay, you can run a small SMTP server on
```
jchate6: I'm not sure how useful this test is. I think the real test most people would want to do is to receive an actual email. Can you give a reason why this should be more convincing that things are working than just writing to stdout?

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+-------------------------------------------------------------------
+What happens when email is not configured or  when sends fail
+-------------------------------------------------------------------
+
+The email features degrade rather than break:
+
+- TOM Toolkit determines whether email is configured with a heuristic predicate
+  (``tom_common.accounts.email.email_is_configured``). Any backend other than Django's SMTP
+  default counts as configured. (The SMTP default pointed at ``localhost`` with no credentials is
+  treated as "not configured".
```
jchate6: again, conservation of parentheses must be maintained! 

I guess technically this open one is balanced by the loose closed one above, but the order is significant.

Again, I don't think a parenthetical is actually required here. That's just useful information. It doesn't need to be a side comment.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+- ``manage.py check`` warns (``tom_common.W002``) when approval-required registration or password
```
jchate6: I'm not convinced these make sense as bullet points.
I think you should re-write this section to just describe what happens when you don't properly configure the email...

You can even add subsections... 

Alternatively, you need each bullet to have a summary word or phrase that let readers know that this is the bullet that they need for their situation.

In general, I think the lends itself better to a FAQ style section than anything else.

"What happens to a registration confirmation?" etc.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+  reset is enabled without a configured email backend.
+- Approving a registration always succeeds even when the notice cannot be sent. Under those
+  circumstances, the approver is told in the UI to notify the user directly. Additionally, the
+  *Pending users* table indicates when email is not configured.
+- A failed password-reset send is reported on the page, with the failure logged for the operator.
```
jchate6: Who is an "operator"? This is another new term. Is it an admin? a manager? a super user? How will they see this log? What will users have to do if their password reset fails?

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,169 @@
...
+
+------------------------
+Customizing the emails
+------------------------
+
+Each email renders from a pair of templates (``*_subject.txt`` and ``*_message.txt``). The
+templates can be customized (overridden) by placing your own copy in your TOM's
+``templates/`` directory (sibling to ``manage.py``). The sender address is ``DEFAULT_FROM_EMAIL``.
+
+- **Registration request** email goes to the addresses in ``MANAGERS``. The default templates are
```
jchate6: I don't think the emphasis is substantial enough here

Image

It still mostly looks like a wall of unorganized text. 

Consider a dash or colon after the emphasized example, and actually use complete sentences after that.
i.e: 

"**Registration request**  - This email goes to the addresses in ``MANAGERS``."

jchate6 (changes requested):
I think some things could be better.

docs/common/email_configuration.rst
phycodurus: The 'django.core.mail.backends.console.EmailBackend' just writes to stdout and there's no SMTP server involved.

This test involves a real SMTP server and relies on more of the configuration
in `settings.py`

docs/common/email_configuration.rst
phycodurus: I see the point, but want to draw the line here as out of scope.

docs/common/email_configuration.rst
```diff
@@ -0,0 +1,191 @@
...
+Configure your ``settings.py`` to point at it (``EMAIL_HOST = 'localhost'``, ``EMAIL_PORT = 8025``),
+and ``sendtestemail`` output should appear in the (`aiosmtpd`) terminal.
+
+-------------------------------------------------------------------
+What happens when email is not configured or  when sends fail
+-------------------------------------------------------------------
+
+Your TOM's Email configuration can be verifiied in code and at the command line:
+
+- TOM Toolkit determines whether email is configured with a heuristic predicate
```
jchate6: 🙄 
I defy you to cleanly explain what you mean by "heuristic predicate" in a way that is useful or informative to the average reader of these docs. 

You literally just need:
"TOM Toolkit determines whether email is configured using ``tom_common.accounts.email.email_is_configured``.

jchate6 (approved):
It feels like you have generally ignored my more important comments about readability, especially where long bullet points are concerned. 

I really don't think multi-line bullets should be included in documentation like this. They undermine the utility of bullets by not allowing users to skim short brief points of information to see what's relevant to them, and they are more difficult to read than complete sentences with surrounding context. That being said, I've made that opinion clear, and they are you're docs.

